PR #11, opened 2025-08-13
Add current directory status item

ropl-btc:
## Summary
Adds a new status item that displays the current working directory name (e.g., "ccstatusline" when in `/path/to/ccstatusline`).

## Changes
- Add `current-dir` status item type
- Implement clean directory name display using `path.basename()`
- Add TUI support with "Current Directory" label
- Update documentation in README

## Implementation
- Follows existing patterns for status items (same structure as `git-branch`)
- Uses yellow as default color to match project styling
- Supports raw value mode (directory name only vs "Dir: name")
- Preview shows "directory" as example text

## Testing
- Tested locally with TUI configuration
- Verified rendering in both preview and live modes
- Confirmed integration with existing status line setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ Once configured, ccstatusline automatically formats your Claude Code status line
 - **Model Name** - Shows the current Claude model (e.g., "Claude 3.5 Sonnet")
 - **Git Branch** - Displays current git branch name
 - **Git Changes** - Shows uncommitted insertions/deletions (e.g., "+42,-10")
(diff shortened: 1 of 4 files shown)


hesreallyhim:
@ropl-btc is it the user's cwd or Claude's? this could be confusing, but I like the idea. IMO it'd be more useful to know Claude's cuz he tends to wander away and get lost.

adelfino69:
yes will be useful

sirmalloc:
I've gone ahead and implemented this myself in v2.0.7.

PR #29, opened 2025-08-18
feat: add Windows support and clean up codebase

Neragoth:
- Add Windows compatibility for terminal width detection
- Remove console.error statements for cleaner output
- Remove DEBUG_FONT_INSTALL debug code
- Remove unnecessary react-devtools-core dependency
- Fix TypeScript/ESLint warnings
- Add .gitattributes for consistent line endings (LF)
- Improve error handling with silent failures

Tested on Windows 11. Unable to test on macOS/Linux platforms.

Capture d'écran 2025-08-18 110320

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Normalize line endings to LF for all text files
+* text=auto eol=lf
+
+# Binary files
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.ttf binary
+*.otf binary
+*.woff binary
+*.woff2 binary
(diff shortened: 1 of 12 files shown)


sirmalloc:
I'll have to review this a bit tomorrow.  I notice you took out the terminal width detection method I was using for POSIX compliant systems.  The reason that was in place is because when claude code runs the statusline as a subprocess, the statusline has no ability to directly access TTY properties, so any standard terminal width detection will always report the width as 80 columns.  The method I had in place works on POSIX systems to find the parent process's TTY and get the actual width using stty.  It's a messy hack, but it's the only option right now to get the full terminal width accurately until Anthropic decides to put it in the JSON passed to the process.

sirmalloc:
Going to close this now as I believe the Windows support has been improved, with the exception of being able to easily infer the terminal width.

PR #50, opened 2025-08-27
chore(biome): Setup biome with lefthook and gh actions

jackall3n:
## Changes

- Replaced eslint with insanely fast biome - I've added some initial ignore rules that match to the existing styling chosen. However, as part of this work I've removed unsused variables etc so there's many code changes. Many of the changes are related to line length and the new lines for each named import.
- Added github action pull request to run lint
- Added lefthook to run biome before each commit, with sub 1s speed

```shell
summary: (done in 0.54 seconds)       
✔️ check (0.53 seconds)
```


sirmalloc:
I'm not really keen on removing my existing stylistic lint rules unless they can be replicated with biome.  I have no problems with a 4s lint time so long as my codebase's consistency in styles is enforced throughout.  There's also strict ordering of imports setup in the current eslint based on my personal preferences.

If you can find a way to preserve all this with biome I'll consider merging this.

jackall3n:
> I'm not really keen on removing my existing stylistic lint rules unless they can be replicated with biome. I have no problems with a 4s lint time so long as my codebase's consistency in styles is enforced throughout. There's also strict ordering of imports setup in the current eslint based on my personal preferences.
> 
> If you can find a way to preserve all this with biome I'll consider merging this.

Sounds fair!

PR #52, opened 2025-08-27
chore(no-git): Hide separators where the previous widget returns null

jackall3n:
## Changes

- Hide separator when widgets return `null` (where `no git` is hidden)


sirmalloc:
Is this for powerline mode or non-powerline mode?

jackall3n:
> Is this for powerline mode or non-powerline mode?

This is for non-powerline currently

sirmalloc:
This will go out whenever I do the next publish.

PR #53, opened 2025-08-27
chore(settings): support more than 3 lines

jackall3n:
## Changes

- Allow user to configure any number of lines
- Added insert and delete controls on line selector screen, with confirmation dialog


sirmalloc:
I'll look at this later today.  Although I'd simplify the example script in package.json and just do a `cat scripts/payload.example.json | bun statusline` without involving the need for a typescript intermediary unless there's some other issue it solves.

jackall3n:
> I'll look at this later today.  Although I'd simplify the example script in package.json and just do a `cat scripts/payload.example.json | bun statusline` without involving the need for a typescript intermediary unless there's some other issue it solves.

Yeah sure. It was to pass the settings path override, but can just cram it all in the package.json script

sirmalloc:
If you have a chance, please update this to use `src/tui/components/ConfirmDialog.tsx` for user consent when deleting a line since it's a destructive operation and could wipe out a good bit of configuration.

sirmalloc:
> Yeah sure. It was to pass the settings path override, but can just cram it all in the package.json script

I see.  I think it'll be sufficient to just do `CCSTATUSLINE_CONFIG=./settings.json bun example` in that case and achieve the same net effect, I believe the environment will pass along to the running script just fine.

sirmalloc:
Also - while the env override is good for testing right now, ultimately I wanted a system where the TUI could offer to save the settings in the project's .claude path or globally, and have the statusline intelligently use the project settings if present.  That would enable them to be checked into source control, while keeping one global ~/.claude/settings.json statusline entry that will use the appropriate settings no matter which project its run in, without the need to launch claude any differently or deal with environment variables.

jackall3n:
> Also - while the env override is good for testing right now, ultimately I wanted a system where the TUI could offer to save the settings in the project's .claude path or globally, and have the statusline intelligently use the project settings if present. That would enable them to be checked into source control, while keeping one global ~/.claude/settings.json statusline entry that will use the appropriate settings no matter which project its run in, without the need to launch claude any differently or deal with environment variables.

Nice, yeah that sounds ideal. I've removed the override for now as I think that's a good way forwards. I've added the confirm dialog, and made some changes to the way the line details are rendered.

PR #54, opened 2025-08-27
feat(active-cwd): show the active working directory of Claude

jackall3n:
## Changes

- Add `CurrentActiveWorkingDir` widget, which displays the latest directory that Claude is in
    - Useful if Claude has moved around a lot and gets confused about where it is

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,6 @@
   ],
   "patchedDependencies": {
     "ink@6.2.0": "patches/ink@6.2.0.patch"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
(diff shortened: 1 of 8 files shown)


jackall3n:
Settings as a draft so that I can understand if this is desired, and will write the tests.

sirmalloc:
This already exists, it's in src/widgets/CurrentWorkingDir.tsx

jackall3n:
> This already exists, it's in src/widgets/CurrentWorkingDir.tsx

I think that's just the cwd from the input json. This is the cwd from the transcript, which changes as Claude navigates the directory

sirmalloc:
This updates as claude navigates, unless you have CLAUDE_BASH_MAINTAIN_PROJECT_WORKING_DIR=1 set in your environment.  The only caveat being that the cwd in the statusline json doesn't update until the message after the cwd was changed (or you can press esc to see it immediately refresh the statusline)

Screenshot 2025-08-27 at 2 19 32 PM
Screenshot 2025-08-27 at 2 19 45 PM

jackall3n:
nice! Not sure how I missed that

PR #72, opened 2025-09-08
Fix cwd segment setting not working as expected on Windows

heromantf:


sirmalloc:
Just pushed this live in 2.0.13, thanks for the contribution!

PR #83, opened 2025-09-18
feat: add remaining mode to context % widgets

jknlsn:
## Changes
- Add remaining/left mode to both Context % widgets
- Press 'l' in widget editor to toggle between used/remaining percentage
- Widgets keep original prefixes, only the percentage value is inverted


sirmalloc:
@jknlsn Thanks for this.  Merged and pushed a new build, v2.0.14

PR #85, opened 2025-09-18
Fix #84: Block timer should reset after idle gap

### LINKED ISSUE #84: Block Timer shows phantom session
## Problem
After a long pause in usage, ccstatusline still reports an active 5‑hour session block. The Block Timer widget continues to show a block that should have ended, so the status line suggests the user is still inside an active window even though usage has already stopped.

## Actual behavior
  - Run ccstatusline with a transcript that includes a gap longer than an hour (e.g., work until roughly 14:22 UTC, then stop until after 18:30 UTC).
  - Open the status line: the Block Timer keeps counting from a start time around 15:00 UTC and displays an ongoing session.
  - Running ccusage blocks --since <date> --until <date> for the same day shows no active block after 14:22 UTC, indicating the session has ended.

## Expected behavior
Block Timer should align with the ccusage blocks report: once ccusage shows no active block, the widget should reset (no remaining block time) instead of projecting a new 5‑hour window starting later in the day.

alex-on-java:
Block windows were staying “active” after multiple hours with no token usage, so the block timer kept counting even though ccusage reported the session closed. Aligning the activity detection with actual token usage prevents those phantom sessions and keeps the timer in sync with the upstream report.

### DIFF
--- a/src/utils/jsonl.ts
+++ b/src/utils/jsonl.ts
@@ -320,13 +320,32 @@ function getAllTimestampsFromFile(filePath: string): Date[] {
 
         for (const line of lines) {
             try {
-                const json = JSON.parse(line) as { timestamp?: string };
-                if (json.timestamp && typeof json.timestamp === 'string') {
-                    const date = new Date(json.timestamp);
-                    if (!Number.isNaN(date.getTime())) {
-                        timestamps.push(date);
-                    }
-                }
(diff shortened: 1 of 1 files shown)


sirmalloc:
@alex-on-java Thanks for this, just merged and published it in v2.0.15

PR #87, opened 2025-09-19
Fix #86 Adjust block metrics gap threshold for short breaks

### LINKED ISSUE #86: Block timer shows longer sessions than daily usage report
When I check today’s blocks in `ccusage`, it shows a new afternoon block starting at 13:00 and running for about ninety minutes. At the same time, the `BlockTimer` widget still claims I have been in the same block for two and a half hours. This mismatch makes it hard to trust the timer, because the widgets are suggesting I worked straight through noon even though `ccusage` clearly split the day into two separate blocks.

alex-on-java:
Detect block boundaries after 60 minutes of inactivity instead of requiring a 5-hour gap. This prevents block start times from carrying over into short midday breaks and ensures elapsed time matches ccusage.

### DIFF
--- a/src/utils/jsonl.ts
+++ b/src/utils/jsonl.ts
@@ -183,6 +183,7 @@ function findMostRecentBlockStartTime(
     sessionDurationHours = 5
 ): BlockMetrics | null {
     const sessionDurationMs = sessionDurationHours * 60 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


sirmalloc:
@alex-on-java I've published this in v2.0.17

PR #89, opened 2025-09-20
Fix #88: Derive block start from rolling hour buckets

### LINKED ISSUE #88: Block Timer drops an hour after consecutive blocks
## Summary
When I finish one five-hour block and roll directly into the next, the Block Timer widget starts the new block an hour late. In the second block for 20 Sep 2025 (20:00:00 → 24:59:59 local time), I had already worked for 1 hour 27 minutes, but the timer only showed 0:27. The `ccusage` report confirmed that I was well into the second block, so the status line lagged by exactly one hour.

## Steps to Reproduce
1. Work continuously through a full block (i.e., accumulate activity from 15:00 to 19:59).  
2. Continue straight into the next block with new messages after 20:00 without taking a long break.  
3. Check the Block Timer during the second block once you’re past the first hour.

## Actual Behavior
- Block Timer shows “0:27” even though I’ve already been active for 1 hour 27 minutes in the second block.  
- `ccusage blocks --json --since 20250920 --until 20250920` reports two consecutive blocks, and the second block absolutely includes 20:00:45–21:27:42 activity.  
- The timer therefore underestimates the elapsed time whenever one block follows immediately after another.

## Expected Behavior
Block Timer should pick up the second block at 20:00 and count from its true start time. If `ccusage` shows a fresh block, the timer should show the correct elapsed time (1:27 in this case) instead of resetting to zero plus one hour of “missing” progress.

alex-on-java:
Earlier we stopped treating a one-hour gap as a hard reset, which solved the phantom “extra” block after mid-session pauses. That change, however, left the block timer one hour behind whenever two five-hour blocks ran back-to-back—the start time snapped to the wrong hour because we were walking the timestamp list as-is.

This update calculates block boundaries exactly the way ccusage does: collect only the hours that actually saw usage, sort those buckets chronologically, and walk backward from the most recent bucket until the window covers five hours. The latest bucket still guards against stale sessions, but the start time now aligns with the first hour in the active rolling window. As a result, mid-session pauses no longer split a block and consecutive blocks no longer drop an hour.

### DIFF
--- a/src/utils/jsonl.ts
+++ b/src/utils/jsonl.ts
@@ -183,7 +183,7 @@ function findMostRecentBlockStartTime(
     sessionDurationHours = 5
 ): BlockMetrics | null {
     const sessionDurationMs = sessionDurationHours * 60 * 60 * 1000;
-    const sessionGapThresholdMs = Math.min(sessionDurationMs, 60 * 60 * 1000); // treat >=1h inactivity as boundary
+    const hourMs = 60 * 60 * 1000;
     const now = new Date();
 
     // Step 1: Find all JSONL files with their modification times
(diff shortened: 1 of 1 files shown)


alex-on-java:
The implementation is incorrect, I'll decline the PR.

PR #98, opened 2025-10-04
Fix context widgets showing 0% due to synthetic API error messages

### LINKED ISSUE #97: Context widgets show 0% when API errors occur in session
## Problem
When a Claude Code session encounters API errors (rate limits, authentication issues, etc.), the context-related widgets incorrectly display 0% or empty values instead of showing the actual context usage. This makes it appear as though no context is being used, even when the session has processed thousands of tokens.

## Actual behavior
- Start a Claude Code session and work normally until context builds up (e.g., 50k+ tokens)
- Encounter an API error (rate limit exceeded, invalid credentials, etc.)
- Check the status line: Context Length, Context %, and Context % (usable) widgets all show 0% or no value
- The session transcript shows the error was appended with all token counts set to 0

Example session:
- Actual context: 81,839 tokens (51% of usable window)
- Displayed context: 0 tokens (0%)

## Expected behavior
Context widgets should display the actual context usage from the most recent real API response, ignoring synthetic error messages. If a session has built up 81k tokens before an error occurs, the status line should still show "Ctx: 81.8k" or "51.1%" rather than "0%".

## Affected widgets
- `src/widgets/ContextLength.ts`
- `src/widgets/ContextPercentage.ts`
- `src/widgets/ContextPercentageUsable.ts`

alex-on-java:
Fix #97 

Investigation revealed that Claude Code now appends synthetic error messages to transcript files when API errors occur (rate limits, authentication failures, etc.). These synthetic messages have:
- model: "<synthetic>"
- isApiErrorMessage: true
- All token usage fields set to 0

The getTokenMetrics() function in src/utils/jsonl.ts calculates the current context length by finding the most recent main chain entry(isSidechain !== true) sorted by timestamp. When an API error occurs, the synthetic error message becomes the "most recent entry", causing contextLength to be calculated as 0.

### DIFF
--- a/src/types/TokenMetrics.ts
+++ b/src/types/TokenMetrics.ts
@@ -9,6 +9,7 @@ export interface TranscriptLine {
     message?: { usage?: TokenUsage };
     isSidechain?: boolean;
     timestamp?: string;
+    isApiErrorMessage?: boolean;
 }
 
 export interface TokenMetrics {
--- a/src/utils/jsonl.ts
+++ b/src/utils/jsonl.ts
@@ -114,7 +114,8 @@ export async function getTokenMetrics(transcriptPath: string): Promise<TokenMetr
                     cachedTokens += data.message.usage.cache_creation_input_tokens ?? 0;
 
                     // Track the most recent entry with isSidechain: false (or undefined, which defaults to main chain)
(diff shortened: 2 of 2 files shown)


sirmalloc:
Thanks, just published in v2.0.20

PR #107, opened 2025-11-03
Fix: Dynamic context window calculation for Sonnet 4.5 (1M tokens)

### LINKED ISSUE #101: Context Percentage -- Sonnet 1m token window
I think, we need to update the code to detect 1m token window (which can be opened /model sonnet[1m]), and use 1m as context, right now it's hardcoded to be 200K?

EthanJStark:
## Summary

Closes #101

This PR fixes incorrect context percentage calculations for Claude Sonnet 4.5, which has a 1M token context window (versus the previously hardcoded 200k).

- Created `model-context` utility with dynamic model-to-context-window mapping
- Sonnet 4.5 models now correctly use 1M context (800k usable at 80%)
- Older models continue using 200k context (160k usable)
- Updated `ContextPercentage` and `ContextPercentageUsable` widgets
- Updated `renderer.ts` flex mode calculations (2 locations)

## Impact

**Before:** Sonnet 4.5 with 42k tokens displayed `21.0%` (42k / 200k)
**After:** Sonnet 4.5 with 42k tokens displays `4.2%` (42k / 1M) ✅

**Older models:** No change in behavior (still 200k denominator)

## Test Plan

- [x] All 36 tests passing (22 existing + 14 new)
- [x] TypeScript type-check passes (`bun run lint`)
- [x] Test coverage for Sonnet 4.5, older models, and edge cases

## Testing Commands

```bash
# Run all tests
bun test

# Lint and type check
bun run lint

# Manual test with Sonnet 4.5
echo '{"model":{"id":"claude-sonnet-4-5-20250929"},"transcript_path":"test.jsonl"}' | bun run src/ccstatusline.ts
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,8 +20,8 @@ bun run start
 # Run directly (TUI mode)
 bun run statusline
 
-# Test with piped input
-echo '{"model":{"id":"claude-sonnet-4-5-20250929"},"transcript_path":"test.jsonl"}' | bun run src/ccstatusline.ts
+# Test with piped input (use [1m] suffix for 1M context models)
+echo '{"model":{"id":"claude-sonnet-4-5-20250929[1m]"},"transcript_path":"test.jsonl"}' | bun run src/ccstatusline.ts
 
 # Build for npm distribution
 bun run build   # Creates dist/ccstatusline.js with Node.js 14+ compatibility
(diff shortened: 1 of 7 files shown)


EthanJStark:
@sirmalloc Review as you're able, thx!

sirmalloc:
From what I can tell, Sonnet 4.5 is only 1M context on API usage. My Max account `/context` shows the limit of 200K still for sonnet 4.5. If I manually do a `/model sonnet[1m]` then `/context` will show 1000K limit, however if I try to make any request it's denied indicating that "the long context beta is not yet available for this subscription".

To do this properly, we would need to look for the [1m] suffix on the model name to safely assume the context limit is 1000K. If we can address these concerns I'm happy to get this merged in quickly.

EthanJStark:
@sirmalloc Aye, great call, _"You're absolutely right!"_ 😉 

942bcca should address this.

Thank you, btw--my team and I all really love ccstatusline!

PR #127, opened 2025-12-11
Fix #126: Use context_window from Claude Code input for context widgets

### LINKED ISSUE #126: Feature request: use `context_window` from Claude Code input when available
# Context widgets should use context_window data from Claude Code input when available

## Problem

Context-related widgets (Context Length, Context %, Context % usable) currently calculate token usage by parsing the transcript JSONL file. This approach has limitations:
- It's slower because it requires reading and parsing the transcript file
- It can be inaccurate when API errors occur (synthetic messages with zero tokens)
- It requires complex model-based context window size lookups

## Background

Claude Code v2.0.65+ now provides a `context_window` object directly in the status JSON input with accurate token metrics:
```json
{
  "context_window": {
    "total_input_tokens": 12345,
    "total_output_tokens": 678,
    "context_window_size": 200000
  }
}
```

This data comes directly from Claude Code and is more reliable than parsing transcript files.

## Expected behavior

When `context_window` data is available in the input (Claude Code v2.0.65+):
1. Context Length widget should use `total_input_tokens` instead of calculating from transcript
2. Context % widget should use `total_input_tokens / context_window_size`
3. Context % (usable) widget should use `total_input_tokens / (context_window_size * 0.8)`

When `context_window` data is not available (older Claude Code versions), widgets should fall back to the existing transcript-based calculation for backward compatibility.

## Affected widgets

- Context Length (`context-length`)
- Context % (`context-percentage`)
- Context % (usable) (`context-percentage-usable`)

alex-on-java:
## Summary

- Use `context_window` data from Claude Code input (v2.0.65+) for context widgets
- Falls back to transcript-based calculation for older Claude Code versions

## Changes

- Added `ContextWindowInfo` interface and `context_window` schema
- Updated `ContextLength`, `ContextPercentage`, and `ContextPercentageUsable` widgets to prefer direct input data
- Maintains full backward compatibility

## Testing

- [x] Linter passes
- [x] All tests pass
- [x] Built successfully
- [x] Manually tested with Claude Code v2.0.65+ input containing `context_window`
- [x] Manually tested with older input format (no `context_window`) - falls back correctly

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -6,7 +6,10 @@ import type {
     BlockMetrics,
     TokenMetrics
 } from './types';
-import type { RenderContext } from './types/RenderContext';
+import type {
+    ContextWindowInfo,
+    RenderContext
+} from './types/RenderContext';
 import type { StatusJSON } from './types/StatusJSON';
 import { StatusJSONSchema } from './types/StatusJSON';
 import { updateColorMap } from './utils/colors';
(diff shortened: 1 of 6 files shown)


alex-on-java:
Need to double check

PR #132, opened 2025-12-25
fix: add Windows compatibility for terminal width detection

### LINKED ISSUE #117: Windows: tput cols creates "null" file and fails terminal width detection
On Windows, the terminal width detection in ccstatusline.js (lines 51447, 51478) runs:

  execSync3("tput cols 2>/dev/null", {...})

  This causes two problems on Windows:

  1. tput is a Unix command that doesn't exist on Windows, producing error: 'tput' is not recognized as an internal or external command
  2. The 2>/dev/null shell redirect is Unix syntax. On Windows CMD, this creates a literal file named null in the current working directory containing the error message.

  Expected behavior: Use cross-platform detection (process.stdout.columns) or skip Unix commands on Windows.

  Suggested fix:
  if (process.platform !== 'win32') {
    // existing tput logic
  }

  Or use process.stdout.columns as the primary method since it's already cross-platform.

  Environment:
  - Windows 11
  - Claude Code with statusLine command

ItsDaRa:
## Summary

- Adds Windows platform check to `getTerminalWidth()` and `canDetectTerminalWidth()` in `src/utils/terminal.ts`
- Uses `process.stdout.columns` on Windows instead of Unix-only shell commands

## Problem

On Windows, the current terminal width detection causes issues:
- `tput` command doesn't exist
- `ps`, `stty`, `awk` are Unix-only utilities
- Shell redirect `2>/dev/null` creates a literal file named "null" in the working directory instead of suppressing stderr

This results in spurious "null" files being created wherever Claude Code is run.

## Solution

Added a platform check at the start of both functions:
```typescript
if (process.platform === 'win32') {
    const cols = process.stdout.columns;
    // return width or null/boolean as appropriate
}
```

`process.stdout.columns` is the Node.js cross-platform API for terminal width and works reliably on Windows.

## Test plan

- [x] Tested on Windows 11 - no more "null" files created
- [x] Terminal width detection works correctly
- [ ] Unix behavior unchanged (early return on Windows means Unix code path untouched)

Fixes #117

### DIFF
--- a/src/utils/terminal.ts
+++ b/src/utils/terminal.ts
@@ -34,13 +34,9 @@ export function getPackageVersion(): string {
 
 // Get terminal width
 export function getTerminalWidth(): number | null {
-    // Windows: use Node.js built-in (cross-platform, avoids Unix-only commands)
-    // This also works on Unix but we keep the existing logic there for compatibility
+    // Preserve historical behavior on Windows: width detection is unavailable.
+    // This avoids Unix fallback command behavior (e.g. 2>/dev/null) on Windows.
     if (process.platform === 'win32') {
-        const cols = process.stdout.columns;
-        if (cols && cols > 0) {
-            return cols;
-        }
         return null;
     }
 
@@ -93,10 +89,9 @@ export function getTerminalWidth(): number | null {
 
 // Check if terminal width detection is available
 export function canDetectTerminalWidth(): boolean {
-    // Windows: use Node.js built-in (cross-platform, avoids Unix-only commands)
+    // Preserve historical behavior on Windows: width detection is unavailable.
     if (process.platform === 'win32') {
-        const cols = process.stdout.columns;
-        return cols !== undefined && cols > 0;
+        return false;
     }
 
     try {


sirmalloc:
@ItsDaRa Have you tested this and seen the terminal width return appropriately on Windows? The main issue has always been that Claude Code calls the statusline as a subprocess without an actual connection to the terminal Claude Code is running in, so typical width detection always returns 80. If you can verify this by adding the "terminal width" widget to your statusline and resizing the terminal and verify it updates properly, then I'll merge this and push a new build out. I don't currently have a Windows machine easily accessible.

ItsDaRa:
Apologizes, I didn't get any sort of notification. I'll try to test it in the next days and get back to you

sirmalloc:
I tweaked this and merged it.  We can't rely on process.stdout.columns to tell us the actual columns on windows when we are invoked as a subprocess of claude code, so for now we disable terminal width detection entirely on windows.  This avoids creating the empty file at c:\dev\null.

ItsDaRa:
All right noted. Sorry I didn't get back to you but I just can't make sense out of it. It seems like it's random, making it harder to troubleshoot, it seem to happen more often when 2 sessions opened in the same repo, but that's all I could get. If I had to bet, I would say it's on Claude Code side, more than ccstatusline itself.

PR #138, opened 2026-01-14
fix: support bare repo worktrees in GitWorktree widget

gn0rt0n:
## Summary

The GitWorktree widget now correctly detects worktrees from bare repositories.

Previously, it only checked for `.git/worktrees/` in the path, but bare repos store worktree metadata at `<bare-repo>/worktrees/<name>` without the `.git` prefix.

**Before:** Worktrees from bare repos showed "no git" instead of the worktree name
**After:** Correctly displays the worktree name

## Changes

- Added check for `/worktrees/` pattern when `.git/worktrees/` isn't found
- Added tests for bare repo worktree detection

## Test plan

- [x] All existing tests pass
- [x] New tests added for bare repo worktrees
- [x] Manually verified with a bare repo worktree setup

## Example

Bare repo setup:
```
/path/to/repo/           <- bare repo (objects/, refs/, HEAD directly here)
├── worktrees/feature-x/ <- worktree metadata
└── trees/feature-x/     <- actual working directory
```

Running `git rev-parse --git-dir` from `trees/feature-x/` returns `/path/to/repo/worktrees/feature-x`, which now correctly extracts `feature-x` as the worktree name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@
 
 ## 🆕 Recent Updates
 
-### v2.1.0 - v2.1.9 - Usage widgets, links, new git insertions / deletions widgets, and reliability fixes
+### v2.1.0 - v2.1.10 - Usage widgets, links, new git insertions / deletions widgets, and reliability fixes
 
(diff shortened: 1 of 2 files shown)


sirmalloc:
Thanks for this, it'll be published as v2.1.10 shortly

PR #142, opened 2026-01-24
Add Session Name widget

arjunlall:
## Summary

Adds a native **Session Name** widget that displays the Claude Code session name set via the `/rename` command.

Currently, displaying session names requires a custom-command workaround with a bash script that parses the session file. This PR adds native support, making it simpler to configure and more performant (direct file read vs spawning a subprocess).

## Implementation

- Reads `transcript_path` from the status JSON (already available in context)
- Parses the JSONL file to find `custom-title` entries
- Returns the most recent session name (searches from end of file)
- Supports `rawValue` mode like other widgets
- Returns `null` when no session name is set (widget hidden)

## Changes

- `src/widgets/SessionName.ts` - New widget implementation
- `src/widgets/__tests__/SessionName.test.ts` - 9 unit tests
- `src/widgets/index.ts` - Export
- `src/utils/widgets.ts` - Registry entry
- `README.md` - Documentation

## Test plan

- [x] All existing tests pass (49 total)
- [x] New widget tests pass (9 tests)
- [x] Lint passes
- [x] Build passes
- [x] Manual testing with test JSONL file

### DIFF
--- a/src/widgets/SessionName.ts
+++ b/src/widgets/SessionName.ts
@@ -12,6 +12,7 @@ export class SessionNameWidget implements Widget {
     getDefaultColor(): string { return 'cyan'; }
     getDescription(): string { return 'Shows the session name set via /rename command in Claude Code'; }
     getDisplayName(): string { return 'Session Name'; }
+    getCategory(): string { return 'Session'; }
     getEditorDisplay(item: WidgetItem): WidgetEditorDisplay {
         return { displayText: this.getDisplayName() };
     }
--- a/src/widgets/__tests__/SessionName.test.ts
+++ b/src/widgets/__tests__/SessionName.test.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs';
 import {
+    afterEach,
     beforeEach,
     describe,
     expect,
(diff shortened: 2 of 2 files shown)


sirmalloc:
Thanks for this, it'll go out in the next release

arjunlall:
I think how claude code handles this may have changed. Ill take a look and file another fix if so. Its also handled natively in CC so less needed in statusline, but a good option nonetheless if it works.

sirmalloc:
> I think how claude code handles this may have changed. Ill take a look and file another fix if so. Its also handled natively in CC so less needed in statusline, but a good option nonetheless if it works.

@arjunlall I tested it briefly with a `/rename` in v2.1.50 of CC and it seemed to pick it up just fine.  Did they add it to the statusline json or something recently?

arjunlall:
Oh great, I was using a custom script as a workaround (until this merged), and that recently broke. Glad the code you tested is still good.

They didn't add it to the statusline, but instead the whole prompt input border turns blue when renamed and shows the label at the upper right of the prompt. It's perhaps redundant in the statusline, but doesn't hurt to give users the option to include it.

PR #150, opened 2026-01-30
feat: add SessionName widget

wenerme:
## Summary

Add a new **SessionName** widget that displays the custom session name set via the `/rename` command in Claude Code.

## Implementation Details

- The widget reads the `sessions-index.json` file from the same directory as the transcript file
- It looks up the `customTitle` field by matching the `session_id`
- Supports raw value mode: displays just the session name or `Session: <name>`
- Returns `null` if no custom title is set (widget won't render)

## Changes

- **New file**: `src/widgets/SessionName.ts` - Widget implementation
- **Modified**: `src/widgets/index.ts` - Export the new widget
- **Modified**: `src/utils/widgets.ts` - Register `session-name` in the widget registry

## Testing

- [x] Lint passes (`bun run lint`)
- [x] TypeScript compiles without errors


sirmalloc:
I've merged an earlier PR #142 that implements this functionality and it's now live in  the most recent release

PR #161, opened 2026-02-03
perf: cache block timer metrics to reduce I/O

### LINKED ISSUE #137: High CPU usage due to full transcript file parsing on every status line render
## Problem

The `ccstatusline` tool causes high CPU usage, especially for users with large transcript files. During profiling with macOS `sample`, the process was observed consuming 60-80% CPU, with the majority of time spent in V8's JSON parser (`v8::internal::Builtin_JsonParse`).

## Root Cause

The `getTokenMetrics()` function (and similar functions like `getSessionDuration()` / `getAllTimestampsFromFile()`) reads and parses the **entire transcript file on every status line render**:

```javascript
async function getTokenMetrics(transcriptPath) {
  // Reads ENTIRE transcript file into memory
  const content = await readFile4(transcriptPath, "utf-8");
  const lines = content.trim().split("\n");
  
  // Parses EVERY line as JSON
  for (const line of lines) {
    const data = JSON.parse(line);  // <-- CPU hotspot
    // ... processes usage data
  }
}
```

This is O(n) where n = number of lines in the transcript, executed on **every status line update**.

## Why This Is Problematic

- **Full file read every render** - The entire transcript file is read into memory on each status line update
- **JSON.parse on every line** - Each line in the transcript is parsed as JSON (profiling showed ~44% of CPU time in JSON parsing)
- **Transcript files grow large** - Long Claude sessions accumulate thousands of JSON lines, each containing full message content
- **Frequent renders** - Status line updates frequently during streaming responses

Users with long-running sessions or large transcripts will experience significant CPU overhead.

## Profiling Evidence

From macOS `sample` output:
```
351 v8::internal::Builtin_JsonParse
  350 v8::internal::JsonParser<unsigned short>::Parse
    314 ParseJsonObject
      95 ParseJsonArray (nested)
```

## Suggested Fixes

| Fix | Description |
|-----|-------------|
| **Cache metrics** | Only re-read file when modification time changes (`fs.stat` mtime check) |
| **Incremental reading** | Track last read byte position, only parse new lines since last read |
| **Tail-only parsing** | For most metrics, only read the last N lines (e.g., using a reverse file reader) |
| **Background worker** | Calculate metrics on an interval (e.g., every 5 seconds) rather than on every render |
| **Lazy widget loading** | Only compute metrics if the relevant widget is actually displayed in the user's config |

## Environment

- **ccstatusline version:** 2.0.23
- **Runtime:** Node.js running via `npm exec ccstatusline@latest`
- **Platform:** macOS

Jaahuo:
Closes #137

## Summary
- Caches block start time to `~/.cache/ccstatusline/block-cache.json` to avoid parsing all JSONL files on every status line render
- Cache is automatically invalidated when the 5-hour block expires
- Falls back to full calculation on cache miss or expiration

## Changes
- **src/ccstatusline.ts**: Changed from `getBlockMetrics` to `getCachedBlockMetrics`
- **src/utils/jsonl.ts**: Added cache functions:
  - `getBlockCachePath()`: Returns path to cache file
  - `readBlockCache()`: Reads cached start time
  - `writeBlockCache()`: Writes start time to cache
  - `getCachedBlockMetrics()`: Main function using cache with fallback
- **src/utils/__tests__/block-cache.test.ts**: Comprehensive test coverage for cache functionality

## Test plan
- [x] All existing tests pass (`bun test` - note: pre-existing GitWorktree.test.ts failure unrelated to this PR)
- [x] New block-cache tests pass (12 tests)
- [x] Lint passes for changed files (`bun run lint` - note: pre-existing lint issues in GitWorktree.test.ts unrelated to this PR)
- [x] TypeScript type check passes

### DIFF
--- a/src/utils/__tests__/block-cache.test.ts
+++ b/src/utils/__tests__/block-cache.test.ts
@@ -1,4 +1,5 @@
 import * as fs from 'fs';
+import { createHash } from 'node:crypto';
 import os from 'node:os';
 import path from 'node:path';
 import {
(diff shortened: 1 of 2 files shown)


sirmalloc:
Made some tweaks to deal with the fact the user can have multiple claude instances using different config dirs via env vars, so cache file is now named with a hash of the config path.  This will be published soon.

Issue #164, opened 2026-02-06
Context percentage for Opus 4.6 1M


zhyu:
Not sure why, but the current implementation hardcoded to only support Sonnet 4.5 1M, which means any new models with 1M context window will not be recognized. Now Opus 4.6 1M is available, an update is needed. If we do not want to update the code every time a new 1M model becomes available, the implementation should be revised.

https://github.com/sirmalloc/ccstatusline/blob/3050aefc1c3412fc9878aff654a7887e5ff660a3/src/utils/model-context.ts#L16-L25

crimist:
#126 should fix and there's a related PR

sirmalloc:
This is fixed with a simple patch in 7cf85dffb12ab08270101a71d385f01eb350a286, however I will review #126 in the future for a more comprehensive fix

PR #160, opened 2026-02-03
feat: use context_window data from Claude Code input for context widgets

Jaahuo:
## Summary
- Extract token metrics from the `context_window` object provided by Claude Code v2.0.65+ instead of parsing transcript JSONL files
- Add `input-parsers.ts` utility with `extractTokenMetricsFromContextWindow()` function
- Update `ContextPercentage` and `ContextPercentageUsable` widgets to use new data source
- Fall back to JSONL parsing when `context_window` is not available for backward compatibility

## Test plan
- [x] Unit tests added for `extractTokenMetricsFromContextWindow()` in `input-parsers.test.ts`
- [x] Updated tests for `ContextPercentage` and `ContextPercentageUsable` widgets
- [x] Manual testing with `bun run example` using updated payload
- [ ] Verify backward compatibility with older Claude Code versions (no `context_window`)

Closes #126

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #140, opened 2026-01-19
[BUG] Statusline always showing 100% usable context and 0 context


Troyanovsky:
Recently, the statusline is always showing 100% usable context and 0 context used. It was fine a about a week ago.

<img width="1055" height="542" alt="Image" src="https://github.com/user-attachments/assets/19d8b622-aa0a-4d54-8e27-a0bd2b29ee83" />

vinzenz:
https://github.com/anthropics/claude-code/issues/18488
Something's broken with Claude Code at the moment even where your context is displayed.

Tbthr:
I encountered the same problem. In version 2.1.11, ctx was displayed normally, but now it remains at 0

dikamilo:
It may be related to GLM models since Anthropic models work well for me and show correct context values, but GLM always shows invalid ones. 

And this is not an issue with thisproject. This is a problem with the claude code or the GLM integration itself, because I tried to create my custom status line and there is the same problem.

AcWrong02:
The issue I’m encountering is that sometimes it shows as 0, and other times it displays normally. The model being used is GLM.

OneChirpZ:
+1
It seems that the usage format returned by the zhipu API (GLM) is inconsistent with that of Anthropic.

PhantomLasagne:
**cross posted from another issue thread**

I'm having similar issues, though, strangely, the context will display correctly anywhere from 20-50% of the time. In my statusline I'm using context length, tokens cached (always displays accurately) and context % (usable). I've attempted a few fixes with changing settings and saving settings with/without CC open with no improvement.

Workspace: 
- VS Code (Win11 install run through WSL2)
- Claude Code CLI
- GLM-4.7 / Minimax-M2.1


Incorrect statusline display:

Model: glm-4.7  |  v2.1.33  |  Style: Explanatory  |                                               
⎇ main  |  𖠰 main  |  (+93,-390)  |                               
Session: 7m  |  Total: 428.7k  |  In: 70.8k  Out: 3.6k  |                                          
Ctx: 0  |  Cached: 354.3k  |  Ctx(u): 100.0%  |


And a perfect example, while typing this the statusline started to display correctly:

Model: glm-4.7  |  v2.1.33  |  Style: Explanatory  |                                               
⎇ main  |  𖠰 main  |  (+154,-441)  | 
Session: 16m  |  Total: 1.0M  |  In: 143.9k  Out: 6.0k  |  
[...]

PR #172, opened 2026-02-12
fix: recognize [1m] suffix on all models, not just Sonnet 4.5

### LINKED ISSUE #171: Context percentage incorrect for non-Sonnet models with [1m] suffix
## Bug Description

`getContextConfig` in `src/utils/model-context.ts` only recognizes `claude-sonnet-4-5` models with the `[1m]` suffix for 1M context. Other models that also support 1M context (e.g. `claude-opus-4-6[1m]`) fall back to the 200k default, causing the context percentage widgets to display incorrect values (~5x too high).

## Steps to Reproduce

1. Set model to `opus[1m]` (or any non-Sonnet model with `[1m]` suffix) in `~/.claude/settings.json`
2. Observe the context percentage in the status line

## Expected Behavior

Context percentage should be calculated against 1M tokens for any model with the `[1m]` suffix.

## Actual Behavior

Context percentage is calculated against 200k tokens, making the displayed usage ~5x higher than actual.

## Root Cause

The condition in `getContextConfig` requires both `claude-sonnet-4-5` in the model ID *and* the `[1m]` suffix:

```typescript
if (
    modelId.includes('claude-sonnet-4-5')
    && modelId.toLowerCase().includes('[1m]')
)
```

The `[1m]` suffix alone is sufficient to indicate 1M context support regardless of model family.

## Fix

Remove the `claude-sonnet-4-5` check and match on the `[1m]` suffix alone. PR incoming.

AniAggarwal:
## Summary

Fixes #171

- `getContextConfig` was hardcoded to only match `claude-sonnet-4-5` models with the `[1m]` suffix
- Other models like `claude-opus-4-6[1m]` also support 1M context but were falling back to the 200k default, causing incorrect context percentage display
- The `[1m]` suffix itself is sufficient to indicate 1M context regardless of model family

## Changes

- **`src/utils/model-context.ts`**: Simplified the condition to check only for the `[1m]` suffix instead of requiring both a specific model name and the suffix
- **`src/utils/__tests__/model-context.test.ts`**: Added test case for `claude-opus-4-6[1m]` and updated test group descriptions to reflect the broader scope

## Test plan

- [x] All 41 existing tests pass
- [x] New test for `claude-opus-4-6[1m]` passes
- [x] Existing Sonnet 4.5 tests still pass (no regression)

### DIFF
--- a/src/utils/__tests__/model-context.test.ts
+++ b/src/utils/__tests__/model-context.test.ts
@@ -7,9 +7,16 @@ import {
 import { getContextConfig } from '../model-context';
 
 describe('getContextConfig', () => {
-    describe('Sonnet 4.5 models with [1m] suffix', () => {
+    describe('Models with [1m] suffix', () => {
         it('should return 1M context window for claude-sonnet-4-5 with [1m] suffix', () => {
             const config = getContextConfig('claude-sonnet-4-5-20250929[1m]');
+
+            expect(config.maxTokens).toBe(1000000);
+            expect(config.usableTokens).toBe(800000);
+        });
+
+        it('should return 1M context window for claude-opus-4-6 with [1m] suffix', () => {
+            const config = getContextConfig('claude-opus-4-6[1m]');
 
             expect(config.maxTokens).toBe(1000000);
             expect(config.usableTokens).toBe(800000);
@@ -32,7 +39,7 @@ describe('getContextConfig', () => {
         });
     });
 
-    describe('Sonnet 4.5 models without [1m] suffix', () => {
+    describe('Models without [1m] suffix', () => {
         it('should return 200k context window for claude-sonnet-4-5 without [1m] suffix', () => {
             const config = getContextConfig('claude-sonnet-4-5-20250929');
 
--- a/src/utils/model-context.ts
+++ b/src/utils/model-context.ts
@@ -13,11 +13,8 @@ export function getContextConfig(modelId?: string): ModelContextConfig {
     if (!modelId)
         return defaultConfig;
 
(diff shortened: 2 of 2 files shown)


EthanJStark:
The [1m] suffix fix is essential and closes #180, #164, and #171. However, I'd like to propose consolidating this with PR #160 into a single, unified effort that addresses all related issues holistically.

### Context

PR #160 adds native `context_window` support from Claude Code v2.0.65+, which is complementary to this fix. Together, they close 7 related issues:
- **#180, #164, #171**: Wrong context % for non-Sonnet 1M models (fixed by this PR)
- **#126**: Use native context_window data (fixed by PR #160)
- **#109, #140, #137**: Context underreporting, CPU usage (both)

### Proposed Approach

I'm implementing a unified plan that:

1. **Task 1**: Merges this PR's [1m] fix (removes the `claude-sonnet-4-5` requirement)
2. **Task 2-4**: Adds native `context_window` support from PR #160 (cleaner architecture than PR #160's approach)
3. **Task 5**: Bonus optimization from PR #160: extracts `formatDurationMs` and uses `cost.total_duration_ms` for the session clock (avoids transcript parsing)
4. **Tasks 6-9**: Complete the context widget updates and documentation

This unified PR will be more cohesive, with a single commit history, consistent test coverage, and better separation of concerns (e.g., `contextWindow` as a separate `RenderContext` field rather than baking percentages into `TokenMetrics`).

### Next Steps

PR #181 is now open on the `fix/1-million-ctx-bug` branch and supersedes this one. Proposing to close both #172 and #160 in favor of that unified approach.

PR #168, opened 2026-02-11
feat: add API usage widgets (session, weekly, reset, context bar)

pcvelz:
## Summary

Adds four new widgets that display Claude Code API usage data, closing #94:

- **session-usage**: 5-hour session utilization with progress bar
- **weekly-usage**: 7-day utilization with progress bar  
- **reset-timer**: countdown to next session reset
- **context-bar**: context window usage with progress bar (using `context_window` data from Claude Code stdin JSON)

Also enhances the existing `ContextPercentage` widget to use `context_window` data when transcript-based token metrics aren't available.

## How it works

- Fetches from `GET /api/oauth/usage` using existing OAuth credentials (macOS Keychain + Linux `~/.claude/.credentials.json` fallback)
- 180-second cache + 30-second rate limit lock to minimize API calls
- Widgets register in the standard widget registry — users add them via the TUI like any other widget
- No changes to default widget layout

## Files changed

| File | Change |
|------|--------|
| `src/widgets/ApiUsage.tsx` | New — all 4 widget classes + credential reading + API fetch + caching |
| `src/widgets/index.ts` | Export new widgets |
| `src/utils/widgets.ts` | Register in widget registry |
| `src/types/StatusJSON.ts` | Add `context_window` schema for Claude Code's context data |
| `src/widgets/ContextPercentage.ts` | Enhanced to use `context_window` from stdin when available |

## Context

This was originally prototyped as a [bash script](https://github.com/sirmalloc/ccstatusline/issues/94#issuecomment-3602367570) and then implemented as native TypeScript widgets in a [fork](https://github.com/pcvelz/ccstatusline-usage). This PR extracts just the widget code with no other changes.

Uses the same API approach as [claude-pulse](https://github.com/NoobyGains/claude-pulse) — same endpoint, same auth headers.

## Test plan

- [ ] `bun run build` succeeds
- [ ] `bun test` passes
- [ ] Widgets appear in TUI widget list and can be added to lines
- [ ] Piped input renders session/weekly usage bars when credentials are available
- [ ] Shows `[No credentials]` gracefully when OAuth token is unavailable
- [ ] Cache file created at `~/.cache/ccstatusline-api.json` after first API call
- [ ] Verify on Linux with `~/.claude/.credentials.json`

PR #181, opened 2026-02-18
feat: native context_window support and generic [1m] suffix handling

### LINKED ISSUE #164: Context percentage for Opus 4.6 1M

### LINKED ISSUE #180: Bug: Sonnet 4.6 with [1m] suffix reports wrong context percentage (uses 200k instead of 1M)
## Description

The 1M context window fix from #107 hardcodes a check for \`claude-sonnet-4-5\` model IDs. With the release of Sonnet 4.6, any \`claude-sonnet-4-6[1m]\` model falls through to the 200k default, causing the context percentage widget to report a grossly inflated percentage.

## Steps to Reproduce

1. Use Claude Code with model \`global.anthropic.claude-sonnet-4-6[1m]\`
2. Observe the ccstatusline context percentage widget

## Expected vs Actual

Claude Code's \`/context\` reports:
```
global.anthropic.claude-sonnet-4-6[1m] · 103k/1000k tokens (10%)
```

ccstatusline reports: **51.3%**

**Root cause:** \`103k / 200k ≈ 51.5%\` — the model is being treated as a 200k context model instead of 1M.

## Root Cause

In \`src/utils/model-context.ts\`, the 1M context check is model-version-specific:

```ts
// Only matches sonnet-4-5, misses sonnet-4-6 and future models
if (
    modelId.includes('claude-sonnet-4-5')
    && modelId.toLowerCase().includes('[1m]')
) {
```

Any model with a \`[1m]\` suffix (indicating long context beta access) should use the 1M context window, regardless of the specific model version.

## Suggested Fix

Replace the version-specific check with a generic \`[1m]\` suffix check:

```ts
// Any model with the [1m] suffix has a 1M context window
if (modelId.toLowerCase().includes('[1m]')) {
    return {
        maxTokens: 1000000,
        usableTokens: 800000 // 80% of 1M
    };
}
```

This would correctly handle Sonnet 4.6, and any future models that support the long context beta.

EthanJStark:
## Summary

- **Fix [1m] suffix check to be model-agnostic** — any model ID ending in `[1m]` now gets 1M context window, not just Sonnet 4.5 (closes #180, closes #164)
- **Native `context_window` support** — when Claude Code v2.0.65+ provides `context_window` in the status JSON, all context widgets use it directly for accurate, real-time metrics instead of parsing the transcript file (partially addresses #126)
- **Session clock optimization** — uses `cost.total_duration_ms` from the status JSON directly when available, avoiding transcript file I/O

## Details

Context widgets (`ContextLength`, `ContextPercentage`, `ContextPercentageUsable`) and `calculateContextPercentage` now check `context_window` first, falling back to transcript-based calculation for Claude Code versions that don't provide it.

`formatDurationMs` is extracted as a shared utility used by both `getSessionDuration` and the new `cost.total_duration_ms` path.

## Test plan

- [ ] All existing tests pass (`bun test`)
- [ ] New model-context tests: Sonnet 4.6, Opus 4.6, Bedrock format, and generic future models with `[1m]` suffix all get 1M context window
- [ ] New widget tests: context widgets use `context_window` data when present, fall back to `tokenMetrics` when absent
- [ ] New `formatDurationMs` unit tests covering `<1m`, minutes-only, hours-only, and hours+minutes cases

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -96,13 +96,13 @@ async function renderMultipleLines(data: StatusJSON) {
     }
 
     // Map context_window from Claude Code input (v2.0.65+)
-    // Prefer current_usage (actual context occupancy) over total_input_tokens (cumulative billing cost)
-    const contextWindow = data.context_window ? {
-        totalInputTokens: data.context_window.current_usage
-            ? (data.context_window.current_usage.input_tokens ?? 0)
-                + (data.context_window.current_usage.cache_creation_input_tokens ?? 0)
-                + (data.context_window.current_usage.cache_read_input_tokens ?? 0)
-            : data.context_window.total_input_tokens ?? 0,
+    // Only use native path when current_usage is present (actual context occupancy).
+    // total_input_tokens is cumulative billing cost and severely undercounts with prompt caching.
+    // Without current_usage, fall through to transcript-based calculation.
+    const contextWindow = data.context_window?.current_usage ? {
+        totalInputTokens: (data.context_window.current_usage.input_tokens ?? 0)
+            + (data.context_window.current_usage.cache_creation_input_tokens ?? 0)
+            + (data.context_window.current_usage.cache_read_input_tokens ?? 0),
         contextWindowSize: data.context_window.context_window_size ?? 0
     } : null;
 


EthanJStark:
Draft momentarily while fixing some bugs

EthanJStark:
Hiya @sirmalloc , thx for the amazing tool! 

Drop me a line whenever you're ready to review this, I'm happy to resolve any conflicts that pop up between now and then.

sirmalloc:
@EthanJStark This had some overlap with PR #168, so I've ensured all this was handled during the merge.  Closing this, will publish a new npm package shortly.

PR #206, opened 2026-03-07
fix: use transcript model to correct cross-session model contamination

### LINKED ISSUE #190: When running multiple worktrees on different models, status line defaults to the most recently changed worktree model for all worktrees
Image

AniAggarwal:
## Summary

Fixes the model display showing the wrong model when multiple Claude Code sessions are open and one changes models. The root cause is upstream (anthropics/claude-code#19570) — the `model` field in statusline JSON reads from global shared state instead of per-session state.

- Extracts the model from the per-session transcript (which has the correct `message.model` on each assistant entry) inside `getTokenMetrics` — no extra file read
- Overrides the stdin model in `renderMultipleLines` when the transcript model differs
- When the last transcript entry is a user message (API call in progress), returns `null` to avoid showing a stale model — self-corrects once the response arrives

### Trade-off

Brief window during API calls where cross-session contamination can still show. Acceptable vs permanently displaying the wrong model. This workaround can be removed when Claude Code fixes the upstream bug.

This is a pretty big patch for a bug in upstream, so feel free to reject. I was getting quite annoyed with the status line changing between chats so figured I'd make a PR.

Closes #190

## Test plan

- [x] 3 new tests: model extraction, pending-user null, missing model null
- [x] All existing tests updated and passing (67 tests across 6 files)
- [x] `tsc --noEmit` clean
- [x] Manual test: two parallel sessions, change model in one, both show correct models

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -38,6 +38,14 @@ import {
 } from './utils/speed-window';
 import { prefetchUsageDataIfNeeded } from './utils/usage-prefetch';
 
+function formatModelDisplayName(modelId: string): string {
+    return modelId
+        .replace(/^claude-/, '')
+        .replace(/-(\d{8})$/, '')
+        .replace(/-/g, ' ')
+        .replace(/\b\w/g, c => c.toUpperCase());
+}
+
 function hasSessionDurationInStatusJson(data: StatusJSON): boolean {
     const durationMs = data.cost?.total_duration_ms;
     return typeof durationMs === 'number' && Number.isFinite(durationMs) && durationMs >= 0;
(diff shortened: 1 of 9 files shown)


sirmalloc:
Thanks for digging into this and putting together a workaround. I looked through the upstream Claude Code issue trail and release notes, and this appears to have been fixed upstream now.

The original bug was tracked as anthropics/claude-code#19570 and deduped into anthropics/claude-code#37596, which is closed as completed. The Claude Code v2.1.86 release notes also explicitly mention fixing the statusline showing another session's model when multiple Claude Code instances are running and `/model` is used in one of them. Related duplicate issues like anthropics/claude-code#35071 and #37884 also confirm the fix landed around 2.1.85/2.1.86.

Given that, I don't think we should carry the transcript-based override here anymore. It was a reasonable workaround while the upstream JSON was unreliable, but now it adds its own edge cases, especially when the latest transcript entry is a user message or when the transcript lags the current session state.

I’m inclined to close this as superseded by the upstream Claude Code fix. Appreciate the PR; it helped clarify the failure mode and gave us a good path while the upstream issue was still open.

AniAggarwal:
Sounds good! 😃

PR #227, opened 2026-03-13
feat: add a global minimalist mode that defaults widgets to raw mode

ctracey:
Adds the foundational plumbing for a global minimalist mode toggle:
- RenderContext.minimalist flag (optional boolean)
- Settings.minimalistMode field (default false, Zod-managed)
- Threads the flag from settings into RenderContext at render time in both piped mode (ccstatusline.ts) and TUI preview (StatusLinePreview.tsx)

Adds an (m) keybind to the Global Overrides TUI menu to toggle minimalist mode on/off. The setting is persisted immediately via onUpdate, and the preview reflects the change in real time via the RenderContext flag.

When minimalist mode is active, the renderer forces rawValue: true on all widgets before calling render(), stripping decorative labels and prefixes globally. No per-widget changes needed.


sirmalloc:

PR #237, opened 2026-03-15
feat: add vim-mode widget

Question406:
## Add Vim Mode widget (`vim-mode`)

Adds a new status line widget that displays the current vim mode when Claude Code's vim mode is enabled.

Example visualization:
image
image

### How it works

Claude Code includes a `vim` field in the status hook JSON when vim mode is active:

```json
{ "vim": { "mode": "NORMAL" } }
```

The widget reads this field and renders a compact mode indicator. When vim mode is not enabled (field absent), the widget returns `null` and hides itself — no empty segment appears.

### Display formats

The widget supports 5 display formats, configurable via `metadata.format` in your settings:

| `metadata.format` | Output | Description |
|---|---|---|
| `icon-dash-letter` *(default)* | `-N` / `-I` | Nerd Font vim icon + dash + letter |
| `icon-letter` | ` N` / ` I` | Icon + space + letter |
| `icon` | `` | Icon only (pair with color) |
| `letter` | `N` / `I` | Single letter |
| `word` | `NORMAL` / `INSERT` | Full word |

Unknown modes fall back to their first character (e.g. `VISUAL` → `V`).

The default `icon-dash-letter` format requires a [Nerd Font](https://www.nerdfonts.com/) — the same requirement already in place for powerline mode.

### Configuration example

```json
{
  "type": "vim-mode",
  "color": "green",
  "metadata": { "format": "icon-dash-letter" }
}
```

### Changes

- **`src/types/StatusJSON.ts`** — added `vim: { mode?: string }` to the status JSON schema
- **`src/widgets/VimMode.ts`** — new widget implementation
- **`src/widgets/__tests__/VimMode.test.ts`** — 10 test cases
- **`src/widgets/index.ts`** — export
- **`src/utils/widget-manifest.ts`** — register `vim-mode` type
- **`src/utils/renderer.ts`** — minor: hoist `getWidget()` call in powerline path (no behaviour change)


sirmalloc:
Thanks for this.  Made a few changes, main one being defaulting to using a non-nerd font glyph by default, and adding a toggle to the editor to enable it, since we can't assume everyone has the font installed.  Also added a key to toggle through the formats in the TUI instead of forcing users to do so in the config file.  I'll publish this in the next release.

PR #245, opened 2026-03-17
fix: discover subagent files by directory scan instead of JSONL references

neoeny152:
## Summary

- **Bug**: `getSubagentTranscriptPaths()` relies on `getReferencedSubagentIds()` finding `agentId` fields in the main session JSONL. However, the main JSONL contains **zero entries with `agentId`** — all agent IDs exist only in the separate subagent files under `{sessionDir}/subagents/`. This causes `getReferencedSubagentIds()` to return an empty set, making `getSubagentTranscriptPaths()` short-circuit and return `[]`, so **no subagent files are ever read**.
- **Impact**: Both speed metrics (tok/s) and token counts miss all subagent tokens. In sessions with heavy agent usage (16+ parallel agents via GSD, team agents, etc.), **90%+ of tokens are invisible** — e.g. showing ~40 tok/s when actual throughput is ~3,333 tok/s.
- **Fix**: Scan `subagents/` directories for all `agent-*.jsonl` files directly instead of requiring pre-discovered IDs. Also add subagent token aggregation to `getTokenMetrics()`.

## Changes

1. **`getSubagentTranscriptPaths()`** — Remove `referencedAgentIds` parameter and the early-return gate. Scan candidate directories for all `agent-*.jsonl` files directly. This is safe because the `subagents/` directory only contains files for the current session.

2. **`getTokenMetrics()`** — After reading the main JSONL, also discover and read all subagent files, summing their token usage (input, output, cached).

3. **`getSpeedMetricsCollection()`** — Remove the broken `getReferencedSubagentIds(mainLines)` call and use the simplified `getSubagentTranscriptPaths(transcriptPath)` directly.

## Verified

Tested against real Claude Code sessions on Windows. Before patch: main JSONL had 0 agent IDs, 0 subagent files discovered. After patch: all subagent files found, ~48% more tokens counted in a 2-agent session. For 16+ agent sessions the difference is ~90%+.

## Test plan

- [ ] Verify speed widgets show combined main+agent tok/s during sessions with subagents
- [ ] Verify token count widgets include subagent tokens
- [ ] Verify backward compatibility when no agents are running (subagents/ dir doesn't exist)
- [ ] Verify no double-counting (main JSONL has `isSidechain: false` only, subagent files have the agent data)

### DIFF
--- a/src/utils/jsonl-metrics.ts
+++ b/src/utils/jsonl-metrics.ts
@@ -194,6 +194,25 @@ export async function getTokenMetrics(transcriptPath: string): Promise<TokenMetr
                 + (usage.cache_creation_input_tokens ?? 0);
         }
 
+        // Include tokens from subagent transcript files
+        const subagentPaths = getSubagentTranscriptPaths(transcriptPath);
+        for (const subagentPath of subagentPaths) {
+            try {
+                const subLines = await readJsonlLines(subagentPath);
+                for (const subLine of subLines) {
+                    const subData = parseJsonlLine(subLine) as TranscriptLine | null;
+                    if (subData?.message?.usage) {
+                        inputTokens += subData.message.usage.input_tokens || 0;
+                        outputTokens += subData.message.usage.output_tokens || 0;
+                        cachedTokens += subData.message.usage.cache_read_input_tokens ?? 0;
+                        cachedTokens += subData.message.usage.cache_creation_input_tokens ?? 0;
+                    }
+                }
+            } catch {
+                // Skip unreadable subagent files
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


sirmalloc:
@neoeny152 I'm not sure this is necessary.  After reviewing my session transcript JSONL files, I've found a 1:1 match of unique agentIds in the JSONL files and the subagent transcripts.  Can you expand on your assertion that the main JSONL has no agentId fields in it, and provide a recent example with the latest Claude Code version?

sirmalloc:
I'm gonna go ahead and close this as I believe it's working as intended with the most recent version of claude code.

PR #252, opened 2026-03-20
fix: detect effort changes from /effort command in ThinkingEffort widget

agolanov:
## Summary
- The ThinkingEffort widget only detected effort changes via `/model` command
  (`Set model to ... with high effort`). The standalone `/effort` command
  (`Set effort level to high: ...`) was not recognized.
- Added detection for the `/effort` command output format in
  `getTranscriptThinkingEffort()`, following the existing prefix + regex +
  normalizeThinkingEffort pattern.
- Added 8 test cases covering all four effort levels, raw value mode,
  priority between `/effort` and `/model`, and settings fallback override.

## Test plan
- [x] All 26 ThinkingEffort tests pass (`bun test src/widgets/__tests__/ThinkingEffort.test.ts`)
- [x] No changes to existing test expectations


orenmizrahi:
this is an important fix since leaving it in effort max (when is says medium) will burn all your tokens in 1 hour instead of 5.

zornot:
+1 — confirmed this fix works as described.

**Environment**: Windows 11, Claude Code 2.1.112, ccstatusline 2.2.8 (installed via `npx ccstatusline@latest`).

**Reproduction before patch**: widget stuck on `Thinking: medium` after running `/effort max` in session, regardless of how many times the statusline refreshed.

**Root cause confirmed by reading the bundled code** (`dist/ccstatusline.js` in v2.2.8): `getTranscriptThinkingEffort()` only recognizes the `/model` stdout pattern (`Set model to ... with X effort`). The `/effort` command produces a different format (`Set effort level to X: ...`) that never matches `MODEL_STDOUT_EFFORT_REGEX`, so the function returns `undefined` and the widget falls back to `settings.json` / the hardcoded `"medium"`.

**After applying the fix from this PR** (manually patched the bundled file): statusline correctly reflects `/effort` changes in real time. Verified with low / medium / high / max transitions on two active sessions.

Would be great to see this merged and published — the workaround right now requires patching `dist/` after every `npx` install.

lanvada:
Thanks for this, @agolanov — this is the missing piece for anyone who uses the built-in `/effort` slash command (as opposed to `/model ... effort ...`). Confirmed still needed against v2.2.8.

The PR is currently marked **CONFLICTING**, almost certainly because #314 (merged 2026-04-17) rewrote parts of `ThinkingEffort.ts` and `jsonl-metadata.ts` to add `xhigh` support. A rebase onto `main` should be mechanical:

- The whitelist in `normalizeThinkingEffort` now includes `xhigh`; the `/effort` branch should emit through the same normalizer so it inherits `xhigh` automatically.
- `MODEL_STDOUT_EFFORT_REGEX` (for `/model`) should stay; add a sibling `EFFORT_STDOUT_EFFORT_REGEX` with prefix `<local-command-stdout>Set effort level to ` and the same `(low|medium|high|xhigh|max)` group — this way both slash commands funnel into one resolver.

Looking forward to this landing in 2.2.9 together with #314. Happy to smoke-test if useful.

sirmalloc:
I'll try to get this reviewed later today and published.

cuipengcx90:
Echoing @lanvada's 4/21 suggestion — sharing the concrete diff that landed cleanly on top of v2.2.10 in my local fork, in case it helps unblock the rebase.

Two micro-changes in `src/utils/jsonl-metadata.ts`:

```ts
const EFFORT_STDOUT_PREFIX = '<local-command-stdout>Set effort level to ';
const EFFORT_STDOUT_REGEX = /^<local-command-stdout>Set effort level to ([a-zA-Z0-9-]+)\b/i;
```

Plus an `EFFORT_STDOUT_PREFIX` branch in `getTranscriptThinkingEffort` that funnels through the same `normalizeThinkingEffort()` already used for `/model`.

Two notes that may help:

- **Why `[a-zA-Z0-9-]+` instead of `(low|medium|high|max)`**: lets `/effort` automatically inherit the `xhigh` whitelist from #314 (and any future levels) with no further regex churn — gating still happens inside `normalizeThinkingEffort`, which is the single source of truth.
- **Plays nicely with #314**: only adds new identifiers + one branch, no overlap with the lines #314 touched, so the rebase should be mechanical.

Verified locally on Opus 4.7 + Claude Code 2.1.x: statusline now reflects `/effort low|medium|high|xhigh|max` transitions in real time, and the existing `/model …` flow is unchanged.

Happy to push a rebase commit if useful, @agolanov / @sirmalloc.

sirmalloc:
@cuipengcx90 Thanks, this is exactly the shape I ended up with when merging main into the PR branch. I kept the `/effort` prefix branch, generalized the regex to `[a-zA-Z0-9-]+`, and routed it through `normalizeThinkingEffort()` so it inherits `xhigh` and future valid levels from the shared parser.

I did hit a small mechanical conflict in `src/utils/jsonl-metadata.ts`, but the resolved version is clean and preserves the existing `/model` and settings fallback behavior.

@agolanov Thanks for this PR, it'll be published in the next release.

PR #253, opened 2026-03-20
feat: use native rate_limits field from Claude Code 2.1.80

Jaahuo:
## Summary

- Claude Code 2.1.80 now sends a `rate_limits` field in the statusline JSON input with `five_hour` and `seven_day` windows containing `used_percentage` and `resets_at` (Unix epoch seconds)
- When `rate_limits` is present, usage widgets consume this data directly, skipping the Anthropic API fetch entirely
- Falls back to the existing API fetch for older Claude Code versions that don't send `rate_limits`
- Updates example payload to match real 2.1.80 format with `context_window`, `rate_limits`, `exceeds_200k_tokens`, and `vim`

## Changes

| File | Change |
|------|--------|
| `src/types/StatusJSON.ts` | Add `rate_limits` to Zod schema with shared `RateLimitPeriodSchema` |
| `src/utils/usage-prefetch.ts` | Add `extractUsageDataFromRateLimits()` with epoch-to-ISO conversion; try `rate_limits` before API fetch |
| `src/ccstatusline.ts` | Pass `data` to `prefetchUsageDataIfNeeded` |
| `src/types/__tests__/StatusJSON.test.ts` | 3 new tests for schema parsing, null, and string coercion |
| `src/utils/__tests__/usage-prefetch.test.ts` | 12 new tests for extraction, fallback, zero percent, null percentage, partial data |
| `scripts/payload.example.json` | Updated to match real 2.1.80 payload |

## Design decisions

- `resets_at` arrives as Unix epoch seconds from `rate_limits` but downstream code expects ISO strings, so conversion happens in `epochSecondsToIsoString`
- `rate_limits` does not include `extra_usage` data — documented in code comment
- No widget changes needed — all usage widgets already consume `UsageData` via `context.usageData`

## Test plan

- [x] All 595 tests pass (`bun test`)
- [x] Type check and lint clean (`bun run lint`)
- [x] Manual test: `cat scripts/payload.example.json | bun start` renders usage widgets with rate_limits data
- [x] Live test: verify usage widgets display correctly with Claude Code 2.1.80

### DIFF
--- a/src/utils/__tests__/usage-prefetch.test.ts
+++ b/src/utils/__tests__/usage-prefetch.test.ts
@@ -126,6 +126,57 @@ describe('usage prefetch', () => {
         expect(usageData).toEqual({ sessionUsage: 42 });
         expect(mockFetchUsageData.mock.calls.length).toBe(1);
     });
+
+    it('falls back to API fetch when seven_day is absent from rate_limits', async () => {
+        mockFetchUsageData.mockResolvedValue({
+            sessionUsage: 50,
+            sessionResetAt: '2026-03-20T12:00:00.000Z',
+            weeklyUsage: 10,
+            weeklyResetAt: '2026-03-27T12:00:00.000Z'
+        });
+
+        const lines = makeLines(
+            [{ id: '1', type: 'weekly-usage' }]
+        );
+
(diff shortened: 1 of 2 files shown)


sirmalloc:
Thanks for this.  Published v2.2.7 with this just now.

PR #265, opened 2026-03-26
fix: detect claude-sonnet-4-6 as 1M context window by default

choobs96:
## Problem

`claude-sonnet-4-6` reports its model identifier without a context size hint (e.g. no `(1M context)` in the display name). This causes `getContextConfig` to fall through to the 200K default, showing inflated context percentages (~5× too high).

Contrast with `claude-opus-4-6`, which reports display name `"Opus 4.6 (1M context)"` — the existing regex catches that correctly.

## Root cause

The `parseContextWindowSize` function relies on the model identifier containing a size hint like `[1m]`, `(1M)`, or `1M context`. For Sonnet 4.5, 1M was an opt-in variant (hence the `[1m]` suffix). For **Sonnet 4.6, 1M is the default** — but the display name Claude Code reports is just `"Claude Sonnet 4.6"`, with no size hint.

## Fix

Add a `DEFAULT_1M_CONTEXT_MODEL_PREFIXES` lookup that catches model identifiers containing `claude-sonnet-4-6` (including AWS Bedrock-prefixed variants) and returns 1M when the regex inference fails.

Explicit `context_window_size` from the status JSON still takes full precedence over this lookup.

## Tests

Three new test cases in `model-context.test.ts`:
- Plain `claude-sonnet-4-6` → 1M ✓
- Bedrock-prefixed `us.anthropic.claude-sonnet-4-6-...-v1:0` → 1M ✓  
- Explicit `contextWindowSize: 200000` still overrides → 200K ✓

All 22 tests pass.

### DIFF
--- a/src/utils/__tests__/model-context.test.ts
+++ b/src/utils/__tests__/model-context.test.ts
@@ -104,6 +104,29 @@ describe('getContextConfig', () => {
         });
     });
 
+    describe('Models with 1M context as default (no suffix required)', () => {
+        it('should return 1M context window for claude-sonnet-4-6', () => {
+            const config = getContextConfig('claude-sonnet-4-6');
+
+            expect(config.maxTokens).toBe(1000000);
+            expect(config.usableTokens).toBe(800000);
+        });
+
+        it('should return 1M context window for claude-sonnet-4-6 with Bedrock prefix', () => {
+            const config = getContextConfig('us.anthropic.claude-sonnet-4-6-20260101-v1:0');
+
+            expect(config.maxTokens).toBe(1000000);
+            expect(config.usableTokens).toBe(800000);
+        });
+
+        it('should still prioritise explicit context_window_size over known-1M-default lookup', () => {
+            const config = getContextConfig('claude-sonnet-4-6', 200000);
+
+            expect(config.maxTokens).toBe(200000);
+            expect(config.usableTokens).toBe(160000);
+        });
+    });
+
     describe('Older/default models', () => {
         it('should return 200k context window for older Sonnet 3.5 model', () => {
             const config = getContextConfig('claude-3-5-sonnet-20241022');
--- a/src/utils/model-context.ts
+++ b/src/utils/model-context.ts
@@ -11,6 +11,13 @@ interface ModelIdentifier {
 const DEFAULT_CONTEXT_WINDOW_SIZE = 200000;
 const USABLE_CONTEXT_RATIO = 0.8;
 
+// Models where 1M context is the default (not an opt-in variant requiring a [1m] suffix).
+// Sonnet 4.6 and later report as plain "claude-sonnet-4-6" with no size hint in the display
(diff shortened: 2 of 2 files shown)


sirmalloc:
@choobs96 Context window length is primarily driven by the `context_window` object in the JSON passed to the statusline, and only uses the model name suffixes as a fallback.  Can you send me the `context_window` object you are receiving while on the default 1M opus?

it should look something like this:

```
  "context_window": {
    "total_input_tokens": 15234,
    "total_output_tokens": 4521,
    "context_window_size": 200000,
    "used_percentage": 8,
    "remaining_percentage": 92,
    "current_usage": {
      "input_tokens": 8500,
      "output_tokens": 1200,
      "cache_creation_input_tokens": 5000,
      "cache_read_input_tokens": 2000
    }
  },
  ```
  
  and `context_window_size` should accurately reflect `1000000` if you are in 1M context mode.  Additionally context percentages are pulled from this object as well, and only fallback to manual calculations in the event this object is not present (older versions of claude code).

**Comment by choobs96** [comment]
Thanks for the pointer! I added debug logging to capture the raw stdin. Here's what Claude Code is actually sending:

```json
{
  "model": {
    "id": "claude-sonnet-4-6",
    "display_name": "Sonnet 4.6"
  },
  "context_window": {
    "context_window_size": 200000,
    "used_percentage": 62,
    "remaining_percentage": 38,
    "current_usage": {
      "cache_read_input_tokens": 124477,
      ...
    }
  }
}
```

So **Claude Code itself is reporting `context_window_size: 200000`** for `claude-sonnet-4-6`, even though the model's actual limit is 1M. The `display_name` is also just `"Sonnet 4.6"` with no `(1M context)` suffix.

This means my current fix is insufficient — the explicit `context_window_size: 200000` takes priority over the model-name lookup, so the override never fires.

The root cause is a bug in how Claude Code reports the context window size for Sonnet 4.6. Would the right fix be to override `context_window_size` when it's ≤ 200K and the model is a known 1M-default model? Something like:

```typescript
if (statusWindowSize !== null) {
    const knownMinSize = getKnown1MDefaultSize(modelIdentifier);
    if (knownMinSize !== null && knownMinSize > statusWindowSize) {
        return { maxTokens: knownMinSize, usableTokens: Math.floor(knownMinSize * USABLE_CONTEXT_RATIO) };
    }
    return { maxTokens: statusWindowSize, usableTokens: Math.floor(statusWindowSize * USABLE_CONTEXT_RATIO) };
}
```

Happy to update the PR if that approach makes sense, or close it if you'd rather wait for the upstream Claude Code fix.

choobs96:
After investigating further, the root cause is actually a bug in Claude Code itself — it's reporting `context_window_size: 200000` for `claude-sonnet-4-6` regardless of the actual 1M limit. Since the fix belongs upstream, I'll close this PR and file a bug with Anthropic instead.

Thanks for the quick response and pointing me in the right direction!

sirmalloc:
@choobs96 That explains a lot, and I suspected that was the case. I'm hoping they will fix this upstream.

PR #278, opened 2026-04-02
Fix token overcounting from streaming duplicate JSONL entries

hgreene624:
## Summary

- Claude Code writes multiple JSONL entries per API call during streaming (intermediate entries have `stop_reason: null`, final entries have `"end_turn"` or `"tool_use"`)
- `getTokenMetrics` was summing **all** entries, inflating the `tokens-total` widget by ~2.5x
- Now only counts entries with a truthy `stop_reason`, with fallback to counting all entries when no `stop_reason` data is present (backward compatibility with older transcripts)

### Before/After (real session, 21 API calls)

| Metric | Before (bug) | After (fix) |
|--------|-------------|-------------|
| JSONL entries counted | 54 | 21 |
| Reported total | 2,427,587 | 957,939 |
| Overcount factor | 2.5x | 1.0x |

## Changes

- **`src/types/TokenMetrics.ts`** — Added `stop_reason` to `TranscriptLine.message`
- **`src/utils/jsonl-metrics.ts`** — Two-pass approach: first scan detects if any entry has `stop_reason`, then second pass skips intermediate streaming entries
- **`src/utils/__tests__/jsonl-metrics.test.ts`** — Added `stopReason` param to test helper, two new tests (streaming dedup + legacy fallback)

## Test plan

- [x] New test: streaming entries are deduplicated, only final entries counted
- [x] New test: legacy transcripts without `stop_reason` fall back to counting all entries
- [x] Existing `jsonl-metrics.test.ts` tests pass (23/23)
- [x] Full suite: no new failures (pre-existing ESM/Bun compat failures in 7 unrelated files)


sirmalloc:
Thanks for this, it will be published in the next release

Issue #92, opened 2025-09-25
Ctx is not being updated after /compact


timurkhakhalev:
Hi! I ran /compact command and it compacted the convo, then I tried to send some message to ensure ccstatusline is updated it's ctx, but it's not. Then I ran /context to see actual ctx size and see that ccstatusline is not updated ctx after the compact.

<img width="1238" height="1035" alt="Image" src="https://github.com/user-attachments/assets/47424e51-b585-4d59-bf04-9a80e108e117" />

sirmalloc:
@timur-khakhalev That's expected, the context in the statusline won't update until you send another message and get a response.  Doing `/compact` or `/context` doesn't send anything back that gets logged to the jsonl that I can parse the context usage off of.  So unfortunately there's no way around this limitation at the moment.

timurkhakhalev:
@sirmalloc I understand that, but I continued the convo and it ctx didn't updated still.

sirmalloc:
@timur-khakhalev I see you have ccusage in the line as a custom command. I know that often takes a while to run, are you sure you waited long enough after getting a response for the status line to update? Did any other field update while the context remained the same?

timurkhakhalev:
Yes, I worked in that convo for another 20 min at least. Didn't notice about update other fields

sirmalloc:
@timur-khakhalev Does this happen as well on any new conversations or just that one?

timurkhakhalev:
> [@timur-khakhalev](https://github.com/timur-khakhalev) Does this happen as well on any new conversations or just that one?

I didn't tried new convo but I think only on this one

PR #282, opened 2026-04-04
feat: add compaction-counter widget

CorticalCode:
## Summary

- Adds a `compaction-counter` widget that tracks context compaction events per session
- When Claude Code compacts conversation context, `used_percentage` drops — this widget detects the drop and displays `↻N`
- Addresses the detection gap noted in #92 by using statusline data instead of JSONL transcripts

## How It Works

Context only grows until compaction — any drop in `used_percentage` > 2 percentage points between consecutive renders is counted as a compaction event. The 2-point threshold filters rounding noise and cache accounting wobble (±1 point) while reliably catching real compaction events on both 200K and 1M context windows.

Per-session state is stored in `~/.cache/ccstatusline/compaction/compaction-{sessionId}.json`, following the skills caching pattern.

### Caveats

- Manual `/compact` detection is delayed until the next assistant message (no statusline update on `/compact` itself)
- `used_percentage` excludes system overhead (~15%, per anthropics/claude-code#34537), but relative drops between renders are still observable
- The 2% threshold is a reasonable starting point; real-world telemetry would help calibrate further

## Changes

| File | Change |
|------|--------|
| `src/utils/compaction.ts` | New — detection logic + session state I/O (Zod-validated) |
| `src/widgets/CompactionCounter.ts` | New — widget class (returns `↻N` or null, supports preview mode) |
| `src/utils/__tests__/compaction.test.ts` | New — 11 detection unit tests |
| `src/widgets/__tests__/CompactionCounter.test.ts` | New — 13 widget tests |
| `src/types/RenderContext.ts` | Added `CompactionData` interface and `compactionData` field |
| `src/ccstatusline.ts` | Compaction detection in render pipeline (conditional, only when widget is configured) |
| `src/utils/widget-manifest.ts` | Registered `compaction-counter` |
| `src/widgets/index.ts` | Barrel export |

## Design Decisions

- **No configurable threshold** — kept simple for first contribution. 2% handles both 200K and 1M windows.
- **Conditional gathering** — compaction detection only runs when the widget is present in the configured layout (follows existing pattern for speed/skills widgets)
- **Zod validation** on state file parse to handle corruption gracefully
- **JSON state files** (not JSONL) since we only track 2 numbers per session

## Test Plan

- [x] `bun test` — 24 new tests pass (11 detection + 13 widget), no regressions
- [x] `bun run lint` — zero errors, zero warnings
- [x] `bun run build` — succeeds, `dist/ccstatusline.js` produced
- [x] 3 pre-existing keychain test failures confirmed present on main (not introduced by this PR)

## Recent changes

  - **Security:** Sanitize session ID before using in cache filename to prevent path traversal
  - **Reliability:** Cache write failures are now best-effort — no longer crash the status line
  - **Tests:** Added persistence round-trip, path traversal regression, and write failure tests

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -13,6 +13,11 @@ import { StatusJSONSchema } from './types/StatusJSON';
 import { getVisibleText } from './utils/ansi';
 import { updateColorMap } from './utils/colors';
 import {
+    detectCompaction,
+    loadCompactionState,
+    saveCompactionState
+} from './utils/compaction';
+import {
     initConfigPath,
     loadSettings,
     saveSettings
@@ -141,6 +146,17 @@ async function renderMultipleLines(data: StatusJSON) {
         skillsMetrics = getSkillsMetrics(data.session_id);
     }
 
+    // Compaction detection — track context percentage drops between renders
+    let compactionCount = 0;
+    const hasCompactionWidget = lines.some(line => line.some(item => item.type === 'compaction-counter'));
+    if (hasCompactionWidget && data.session_id && data.context_window?.used_percentage !== null && data.context_window?.used_percentage !== undefined) {
+        const prevState = loadCompactionState(data.session_id);
+        const ctxPct = Math.round(data.context_window.used_percentage);
+        const newState = detectCompaction(ctxPct, prevState);
+        saveCompactionState(data.session_id, newState);
+        compactionCount = newState.count;
+    }
+
     // Create render context
     const context: RenderContext = {
         data,
@@ -150,6 +166,7 @@ async function renderMultipleLines(data: StatusJSON) {
         usageData,
         sessionDuration,
         skillsMetrics,
+        compactionData: compactionCount > 0 ? { count: compactionCount } : null,
         isPreview: false,
         minimalist: settings.minimalistMode
     };
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -49,6 +49,13 @@ function hasSessionDurationInStatusJson(data: StatusJSON): boolean {
     return typeof durationMs === 'number' && Number.isFinite(durationMs) && durationMs >= 0;
 }
 
+function hasContextPercentage(
+    data: StatusJSON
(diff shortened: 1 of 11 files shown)


CorticalCode:
After closing  PR #309, figured I'd do a proper cleaning run on this one.

Force-pushed: rebase on main + thorough self-review. 4 commits, 988 tests pass, lint clean. Commit message covers what changed.

**Verified beyond unit tests:**

- `bun run build` + bundled output under Node.js
- End-to-end pipe test: 6 scenarios through the real pipeline (cache I/O included)
- TUI manifest, metadata, and render scenarios
- Live session in isolated `CLAUDE_CONFIG_DIR` — ↻N increments on real `/compact`

**Incidental finding (not this widget):**

Layouts with explicit `{type:"separator"}` items leave an empty slot when a hide-capable widget hides (`count=0`) — affects any hide-capable widget.

Root cause: the renderer's `hasContentBefore` check at `src/utils/renderer.ts:670-685` walks all the way back rather than stopping at the immediate prior widget.

Workarounds: position at a line edge, or use `defaultSeparator`.

Looking into options; considering a fresh PR.

**FYI on the counter:**

Back-to-back `/compact` doesn't always increment — if the post-compact drop is below the 2% threshold, the detector ignores it. In practice the second compact is mostly a no-op anyway (nothing left to compact), so this feels semantically right.

sirmalloc:
Thanks.  I made some tweaks, you can see the notes on the latest commit above.  I'll publish this when I do the next release.

CorticalCode:
Appreciate the feedback.

On the window-size baseline: I don't switch models often, but right after submitting the last amended PR, I noticed the counter firing when I changed models. I should have considered that context window size changes would cause a false positive. You caught it right away, and your fix is better than what I was mapping out.

The `calculateContextPercentageMetrics` extraction, format cycling, Nerd Font support, and hide-when-zero toggle all point at the same lesson. I pulled the compaction counter from my status line bash script and submitted it as an isolated widget without really thinking through how to fit in with the rest of the project. I'll aim to wire into existing patterns rather than bolt on from the side on future contributions.

PR #309, opened 2026-04-15
feat: add settings.local.json support for Claude Code integration

### LINKED ISSUE #276: Allow claude.local.settings file to configure ccstatusline
At my company, the `.claude/settings.json` is override for an external tool to configure some global company settings. So, I tried to configure statusline at `.claude/settings.local.json` but it isn't working. I have to manually add each time this settings overrided.

The idea is ccstatusline can check this file to configure properly, not just `.claude/settings.json`.

Thanks in advance for this great tool. 🙂

CorticalCode:
## Summary

Adds `settings.local.json` support to the Claude Code integration layer. Fixes #276.

- When `settings.local.json` exists, the TUI install flow lets users choose which file to target
- Detection functions check both files, matching Claude Code's merge precedence (local wins)
- Uninstall cleans both files but only removes entries ccstatusline owns (`isKnownCommand` guard)
- Hook sync targets the file where the statusLine lives, cleans stale hooks from the other

## Design decisions

1. **Auto-detect:** If `settings.local.json` exists on disk, prompt the user to choose a target. If it doesn't exist, default behavior is unchanged — zero friction for the common case.
2. **Local wins:** `settings.local.json` takes precedence when checking installation status, matching Claude Code's own merge semantics.
3. **Ownership guard:** Uninstall only removes `statusLine` entries identified by `isKnownCommand()`. A company-managed or user-custom statusLine in either file is never touched.
4. **Accurate display:** Confirmation dialogs show the actual file being targeted. A warning appears if installing to `settings.json` while `settings.local.json` already has a statusLine that would take precedence.

## Files changed

- `src/utils/claude-settings.ts` — `filePath` param on load/save, private `resolveStatusLine()` for dual-file precedence, ownership-guarded uninstall
- `src/utils/hooks.ts` — file-aware hook sync targeting active file, stale hook cleanup from non-active file
- `src/tui/components/FileTargetSelect.tsx` — new component for install target selection
- `src/tui/App.tsx` — install flow integration, fileTarget screen, confirmation warnings
- `src/tui/components/InstallMenu.tsx` — `localSettingsDetected` prop for contextual help text
- Tests for all new behavior including path traversal, malformed files, ownership preservation

## Known limitations

- If `settings.json` is read-only (e.g., company-managed with restrictive permissions), write operations that target it will error. The `settings.local.json` support provides a workaround.
- Related but not in scope: PR #58 by @jackall3n (project-level ccstatusline settings) addresses a different settings layer.

## Test plan

- [ ] Install to `settings.json` (default, no `settings.local.json` present) — unchanged behavior
- [ ] Install when `settings.local.json` exists — FileTargetSelect screen appears
- [ ] Install to `settings.local.json` — writes only to local file
- [ ] Uninstall with entries in both files — removes ccstatusline entries, preserves non-ccstatusline entries
- [ ] Hook sync after install to local file — hooks written to local, cleaned from global


sirmalloc:
@CorticalCode I was about to merge this PR and noticed the official docs at https://code.claude.com/docs/en/settings indicate the location of `settings.local.json` is project local and is meant to be found in `.claude/settings.local.json` in the project folder.  I don't think claude ever looks for this file in `~/.claude/settings.local.json`.

When you tested this, was claude respecting the `settings.local.json` found in `~/.claude/`?

CorticalCode:
You're right, took another look. Set up a sentinel: added `echo SENTINEL_FROM_LOCAL_SETTINGS_FILE` to a `statusLine` in `~/.claude/settings.local.json`, left the existing one in `~/.claude/settings.json`, fired up a fresh `claude`. Status line showed the global one. Doesn't work.

Went back and re-read #276 with that in mind. Adrián was talking about project-local `.claude/settings.local.json`; my take was focused on user-local `~/.claude/settings.local.json` different file entirely. And even if Claude did honor user-local, it still wouldn't fix his situation, since the company tooling is overwriting the project-shared file and the override path he actually needs is the project-local one. So the PR was borked from the premise on up.

Sorry for the time you spent on this. I had tunnel vision and didn't vet any of it properly. Closing it out.

There may be a path to a redesign using the existing plumbing (`FileTargetSelect`, dual-file resolution, ownership-guarded uninstall all carry over), but it'd be a rescoped expansion of ccstatusline from "install once at user scope" to "per-project install." Having dumped this turd in your yard, I won't blame you for not wanting me near it again, but if you're open to a second chance, I can try again and at least let you know.

sirmalloc:
@CorticalCode All good.  Feel free to take another stab at it.  The main thing to get right is the UX.  It would need to support global install / uninstall as well as project local install / uninstall.  It should show the status for both in an intuitive way.  And it needs to handle the hooks cleanly when hook-enabled widgets are present.  I haven't had time to give it proper thought to satisfy all this in a clean way so I've been avoiding it.  If you can think of something that works well and I'm happy with it I'll merge it.

PR #311, opened 2026-04-16
fix: context percentage labels reflect used/remaining state

huangguang1999:
## Summary

The `Context %` and `Context % (usable)` widgets currently render a static `Ctx: ` / `Ctx(u): ` label regardless of whether the value shown is *used* or *remaining* (via the `(u)` keybind that toggles `inverse`). So a user glancing at `Ctx: 9.3%` can't tell whether that's 9.3% used or 9.3% remaining without opening the editor to check the inverse modifier.

This PR makes the label reflect the current state:

- default (used):   `Ctx Used: X%`   /  `Ctx(u) Used: X%`
- inverse (left):   `Ctx Left: X%`   /  `Ctx(u) Left: X%`

The existing keybind label `(u)sed/remaining` and the editor modifier text `(remaining)` already communicate these two states; this change just surfaces the same distinction in the rendered output. The `Ctx` / `Ctx(u)` prefix is kept so the two widgets remain visually distinguishable from one another and from `Context Length`'s `Ctx: 18.6k` numeric label.

This is a small UX polish — no behavior change beyond the label string.

## Test plan

- [x] `bun test` — 731 / 731 pass (including the updated `ContextPercentage.test.ts` / `ContextPercentageUsable.test.ts`)
- [x] `bun run lint` — clean
- [x] Manual: run `ccstatusline setup`, add both Context % widgets, toggle each one via `(u)`, confirm the rendered label flips between `Used` and `Left`

## Notes

Happy to adjust the exact label wording (e.g. `Rem:` instead of `Left:`) if you'd prefer something shorter — I went with `Used`/`Left` because they match the mental model of "how much is used" vs "how much is left" and are symmetric in length.


sirmalloc:
Thanks for this, I'll publish it in the next release.

PR #314, opened 2026-04-16
fix: support xhigh thinking effort level

spencersmith:
## Summary
- Adds support for Claude Code's new `xhigh` thinking effort level. Accepts both `xhigh` and `xHigh` casings from `/model` transcript output and from `effortLevel` in `settings.json`.
- Opens the `ThinkingEffort` widget up to future effort levels: any unknown but word-shaped effort name (2–20 chars, `[a-z0-9-]`, case-insensitive) is rendered verbatim with a trailing `?` (e.g. `super-max?`) instead of being silently dropped.
- Changes the `ThinkingEffort` widget's fallback display from a silent `medium` assumption to an explicit `default` when no effort signal is available from either the transcript or Claude Code settings. This avoids showing a misleading effort label when ccstatusline genuinely can't determine the current value (fresh session, no `effortLevel` set, no `/model` command run yet).

## Test plan
- [x] `bun test` — all 742 tests pass, including 11 new cases covering `xhigh`/`xHigh`, `super-max`/`Super-Max` unknown-but-valid values, and rejected shapes (oversized, single char, contains whitespace).
- [x] `bun run lint` — clean.
- [x] Manual verification with piped payloads for each level, the unknown-with-marker path, and the `default` fallback.
- [x] Live verification in a sandboxed Claude Code session with the locally built `dist/ccstatusline.js`.

### DIFF
--- a/src/types/ClaudeSettings.ts
+++ b/src/types/ClaudeSettings.ts
@@ -1,5 +1,7 @@
+import type { TranscriptThinkingEffort } from '../utils/jsonl-metadata';
+
 export interface ClaudeSettings {
-    effortLevel?: 'low' | 'medium' | 'high' | 'max';
+    effortLevel?: TranscriptThinkingEffort;
     permissions?: {
         allow?: string[];
         deny?: string[];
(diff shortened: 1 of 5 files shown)


sirmalloc:
Thanks, will publish this in the next release.

PR #331, opened 2026-04-22
fix: detect terminal width on Windows via parent console

jcmecham:
## Summary

On Windows, `getTerminalWidth()` always returned `null`, so flex separators collapsed and long-line truncation was disabled. Claude Code spawns the statusline with piped stdio and no controlling PTY, so `process.stdout.columns` is undefined; there's no `columns` field in the statusline JSON and no `COLUMNS` env var to fall back on; and the existing Unix `ps`/`stty` probe doesn't apply on Windows.

This PR makes width detection work on Windows by walking up the process tree to the `claude.exe` ancestor (which owns the real console), attaching to that console via a PowerShell P/Invoke helper, and reading the width from the Win32 console API.

#### Performance:
 The probe runs synchronously inside ccstatusline (not on Claude Code's main thread, so the REPL is unaffected). Cold call: ~5s (PowerShell startup + `Add-Type` JIT). The result is cached at `%TEMP%/ccstatusline-win-width-cache.json` with a 60s TTL, so subsequent status-line renders read from disk in ~5ms. Trade-off: up to 60s of resize lag before the cache refreshes — acceptable for a status line.

## Changes

| File | Change |
|------|--------|
| `src/utils/terminal.ts` | Windows branch added to `getTerminalWidth()`. Shells out to the probe, caches the result at `%TEMP%/ccstatusline-win-width-<ppid>.json` with a 60s TTL, returns `null` on any failure (no regression). In TUI mode (stdout is a real TTY) it returns `process.stdout.columns` directly and skips the probe. Non-Windows paths untouched. |
| `scripts/windows-width-probe.ps1` *(new)* | PowerShell + C# P/Invoke probe. Walks PPIDs with `CreateToolhelp32Snapshot`, `AttachConsole`s to the owning process, reads `CONSOLE_SCREEN_BUFFER_INFO.dwSize.X`. Invoked via `spawnSync -EncodedCommand` to avoid `cmd.exe` arg truncation. |
| `src/tui/components/ItemsEditor.tsx` | Removed the "Terminal width detection is currently unavailable in your environment" warning — it existed to explain the broken Windows behavior this PR resolves. |
| `src/utils/__tests__/terminal.test.ts` | New tests for probe parsing, failure modes, and cache-hit behavior. Existing Unix tests now mock `process.platform` explicitly so they pass on Windows dev boxes. |
| `package.json` | `build` script copies the probe into `dist/` so the published package ships it. |

## Test plan

- [x] `bun run vitest run src/utils/__tests__/terminal.test.ts` — 12 tests pass.
- [x] `bun run lint` — clean for touched files.
- [x] Manual: `bun run build`, point `statusLine.command` at `dist/ccstatusline.js`, add a Flex Separator, resize Windows Terminal — line spans full width, reflows on resize within the 60s TTL.
- [x] Manual: open the Items Editor in the TUI and hold down-arrow — instant response, no multi-second lag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "start": "bun run src/ccstatusline.ts",
-    "build": "rm -rf dist/* ; bun build src/ccstatusline.ts --target=node --outfile=dist/ccstatusline.js --target-version=14",
+    "build": "rm -rf dist/* ; bun build src/ccstatusline.ts --target=node --outfile=dist/ccstatusline.js --target-version=14 && cp scripts/windows-width-probe.ps1 dist/",
     "postbuild": "bun run scripts/replace-version.ts",
     "example": "cat scripts/payload.example.json | bun start",
     "prepublishOnly": "bun run build",
--- a/scripts/windows-width-probe.ps1
+++ b/scripts/windows-width-probe.ps1
@@ -1,0 +1,70 @@
+# Walk ancestors looking for claude.exe, attach to its console, emit width.
+# Emits nothing if claude.exe isn't in the chain — caller falls back to null.
+
+$ErrorActionPreference = 'SilentlyContinue'
+
+# @'...'@ is a literal here-string — PowerShell won't interpolate $ inside.
+Add-Type -TypeDefinition @'
+using System;
+using System.Runtime.InteropServices;
+
+public class NP {
+    // Structs matching Win32 memory layout.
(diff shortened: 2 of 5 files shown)


jcmecham:
Verified it works in windows terminal application
TerminalWidthCMD
Verified it works in Intellij

Verified it works in vscode

comertial:
i confirm that flex seperator in windows is not working.. a fix like this must be introduced

PR #332, opened 2026-04-23
feat(tui): add wrap-around navigation to all menus and move modes

iamvkosarev:
## Summary
- Enable circular cursor movement across all TUI menus and lists — pressing up at the first item wraps to the last, and pressing down at the last wraps to the first
- Also applies to move/reorder modes, allowing items to be moved cyclically through list boundaries
- No existing PR covers this functionality

## Changes
- `List.tsx` — change `wrapNavigation` default from `false` to `true`, enabling wrap-around for all `<List>`-based menus     (MainMenu, InstallMenu, PowerlineSetup, etc.)
- `input-handlers.ts` — replace `Math.max`/`Math.min` clamping with wrap-around logic in `handleNormalInputMode`, `handleMoveInputMode`, and `handlePickerInputMode`
- `PowerlineSeparatorEditor.tsx` — wrap-around for separator list navigation
- `LineSelector.tsx` — wrap-around in line move mode

## Test Plan
- 789 tests pass (10 new), TypeScript and ESLint clean
- New tests cover wrap boundaries for: normal mode (up/down), move mode (up/down), picker categories (up/down), picker widgets (up/down), and top-level search (up/down)

### DIFF
--- a/src/tui/components/LineSelector.tsx
+++ b/src/tui/components/LineSelector.tsx
@@ -97,26 +97,28 @@ const LineSelector: React.FC<LineSelectorProps> = ({
         }
 
         if (moveMode) {
-            if (key.upArrow && selectedIndex > 0) {
+            if (key.upArrow && localLines.length > 1) {
                 const newLines = [...localLines];
+                const targetIndex = selectedIndex - 1 < 0 ? localLines.length - 1 : selectedIndex - 1;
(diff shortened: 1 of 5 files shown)


sirmalloc:
Nice!  I like it.  I'll publish it in the next release.

PR #338, opened 2026-04-26
fix(ThinkingEffort): show xhigh effort correctly when only settings.json available

### LINKED ISSUE #336: Thinking effort widget always shows 'medium' for 'xhigh' effort
## Problem
The thinking effort widget consistently displays 'medium' even when the actual Claude thinking effort value is 'xhigh'. This is shown in the screenshot where the effort is set to 'xhigh' but ccstatusline reports 'medium'.

## Proposed Solution
Investigate how the thinking effort value is read and mapped to display labels. The fix should ensure the widget correctly reflects the actual effort level (e.g., 'xhigh', 'high', 'medium', 'low', 'off').

## Use Case
Users who rely on the thinking effort widget to monitor Claude Code's current reasoning level need accurate information to understand the model's behavior during tasks.

fuleinist:
## Problem

The Thinking Effort widget consistently shows 'medium' even when Claude Code's effort level is set to 'xhigh'. Users see 'Thinking: medium' instead of the correct effort level.

## Root Cause

Two issues combined:

1. **settings.json fallback returned  instead of **: When  was undefined,  returned , causing the  fallback to still return  — which then propagated up and eventually hit  returning 'default'.

2. **Transcript path took precedence with no effort line**: When the transcript exists but contains no effort line,  returns  (not null), causing the  chain to still prefer the undefined transcript path over a valid settings fallback.

## Fix

1.  now returns  explicitly when settings are unavailable or have no  — removing the  ambiguity.
2.  now uses explicit if-checks instead of  so transcript takes precedence **only when it actually returns a result** (not when it returns undefined), ensuring the settings fallback is consulted when the transcript has no effort info.

## Testing

- Added test coverage for the transcript-first behavior
- Verified existing tests still pass (xhigh, settings fallback, mixed-case effort)

Closes #336

### DIFF
--- a/src/widgets/ThinkingEffort.ts
+++ b/src/widgets/ThinkingEffort.ts
@@ -15,21 +15,30 @@ import {
 
 export type ThinkingEffortLevel = TranscriptThinkingEffort;
 
-function resolveThinkingEffortFromSettings(): ResolvedThinkingEffort | undefined {
+
+function resolveThinkingEffortFromSettings(): ResolvedThinkingEffort | null {
     try {
         const settings = loadClaudeSettingsSync({ logErrors: false });
-        return normalizeThinkingEffort(settings.effortLevel);
+        if (!settings) {
+            return null;
+        }
+        const resolved = normalizeThinkingEffort(settings.effortLevel);
+        if (resolved) {
+            return resolved;
+        }
     } catch {
-        // Settings unavailable, return undefined
+        // Settings unavailable, return null to trigger fallback
     }
 
-    return undefined;
+    return null;
 }
 
 function resolveThinkingEffort(context: RenderContext): ResolvedThinkingEffort | null {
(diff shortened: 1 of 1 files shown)


sirmalloc:
Thanks for the PR. I took a fresh look against current `main`, and I think this has been superseded.

`main` now handles `xhigh` through the shared thinking-effort normalization, reads live status JSON `effort.level` before transcript/settings fallbacks, and also has `/effort` transcript parsing from #252. The specific fallback behavior described here is already covered by tests on `main` for missing transcript effort, unavailable transcript files, and `xhigh` from `settings.json`.

I also checked the merge shape, and this now conflicts in `src/widgets/ThinkingEffort.ts` without adding behavior that `main` does not already have. Closing this as superseded, but I appreciate the investigation and the clear writeup.

PR #339, opened 2026-04-26
fix(widget): claude-account-email respects CLAUDE_CONFIG_DIR

### LINKED ISSUE #317: claude-account-email widget ignores CLAUDE_CONFIG_DIR when multiple profiles are active
I set up multiple Claude Code profiles using `CLAUDE_CONFIG_DIR` (one for work, one for personal) using aliases:

```bash
alias cc-work='CLAUDE_CONFIG_DIR=$HOME/.claude-work command claude'
alias cc-personal='CLAUDE_CONFIG_DIR=$HOME/.claude-personal command claude'
```
Each profile has its own `.claude.json` with separate auth, and `CLAUDE_CONFIG_DIR` is confirmed set correctly inside sessions (I asked Claude to run `echo $CLAUDE_CONFIG_DIR` in each one).

Even so, the claude-account-email widget shows the same account email in both sessions.  It seems like the widget is reading `~/.claude.json` at the home level rather than `$CLAUDE_CONFIG_DIR/.claude.json`.  I tried removing the base home file, and both sessions no longer show an email account at all.  When I restore it, they go back to showing the same email address.  I know they're using their respective accounts, because Claude Code reports one as a Pro account and the other as Team.

I'm happy to test a fix or provide more information about my setup if it's useful.

CorticalCode:
Closes #317.

The widget reads the same `.claude.json` no matter what `CLAUDE_CONFIG_DIR` is set to, so every profile ends up showing the same account email.

The path was built like this:

```ts
path.join(getClaudeConfigDir(), '..', '.claude.json')
```

That works for the default `~/.claude` config dir, since going up one level lands at `$HOME` where `.claude.json` actually lives. But when someone sets `CLAUDE_CONFIG_DIR=~/.claude-work`, the `..` still lands at `$HOME`, so every profile keeps reading the same file.

The reason it's not a simple fix: `.claude.json` lives at `$HOME` by default, but Claude Code moves it inside `CLAUDE_CONFIG_DIR` when that env var is set. The path needs to branch on whether the env var is there:

```ts
function getClaudeJsonPath(): string {
    const envConfigDir = process.env.CLAUDE_CONFIG_DIR;
    if (envConfigDir) {
        return path.resolve(envConfigDir, '.claude.json');
    }
    return path.join(os.homedir(), '.claude.json');
}
```

A few small extra changes I made in the same render method:

- Removed an `existsSync` check before `readFileSync`. The outer try/catch already handles missing files, and dropping the check also closes a small race window between the two calls.
- Dropped a redundant `path.resolve` on a path that was already absolute.
- Tightened the email check from `if (!email)` to `typeof email === 'string' && email.length > 0`. The old check would let a non-string `emailAddress` field through, so `{"emailAddress": 12345}` would have rendered as `Account: 12345`.

## Tests

Three new tests:

1. Widget reads from `$CLAUDE_CONFIG_DIR/.claude.json` when the env var is set. This is the regression test.
2. Widget returns null when `CLAUDE_CONFIG_DIR` points at a directory with no `.claude.json`, even if `$HOME/.claude.json` exists. Guards against the old fallback coming back.
3. Non-string `emailAddress` returns null.

All existing tests still pass. `bun run lint`, `bun test`, and `bun run build` are all clean.

@daren-porter — if you have a minute, could you check this against your work/personal alias setup? The new tests mirror what you described in the issue, but a real-world confirmation would help.

### DIFF
--- a/src/utils/__tests__/claude-settings.test.ts
+++ b/src/utils/__tests__/claude-settings.test.ts
@@ -16,6 +16,7 @@ import { DEFAULT_SETTINGS } from '../../types/Settings';
 import {
     CCSTATUSLINE_COMMANDS,
     getClaudeCodeVersion,
+    getClaudeJsonPath,
     getClaudeSettingsPath,
     getExistingStatusLine,
     getRefreshInterval,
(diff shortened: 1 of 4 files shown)


sirmalloc:
Thanks.  I made one small tweak the centralize the logic to grab the json path in case we need to use it outside that widget in the future.

PR #220, opened 2026-03-12
feat: support absolute datetime for reset timers

fanyu:
## Summary
- add a date mode (`d`) to both Block Reset Timer and Weekly Reset Timer
- in date mode, render the reset point as an absolute UTC timestamp instead of a duration
- keep existing duration/progress modes unchanged and gracefully fall back to duration if reset timestamp is unavailable

## UX
- `p`: cycle time/full bar/short bar (existing)
- `s`: compact formatting for duration/date (existing behavior extended to date mode)
- `d`: toggle absolute datetime mode (new)

## Implementation
- add `formatUsageResetAt(resetAt, compact)` utility
- add shared metadata helpers `isUsageDateMode` / `toggleUsageDateMode`
- wire date mode into both reset timer widgets and editor modifier text
- update README docs for widget behavior and keybinds
- add/extend tests for date mode rendering and reset timestamp formatter

## Verification
- `npx tsc --noEmit && npx eslint . --config eslint.config.js --max-warnings=0`
- `npx vitest run src/utils/__tests__/usage.test.ts src/widgets/__tests__/BlockResetTimer.test.ts src/widgets/__tests__/WeeklyResetTimer.test.ts`

PR #340, opened 2026-04-27
feat: support local/IANA timezone in reset timer date mode

### LINKED ISSUE #337: Feature Request: Local timezone support for reset timer date mode (#220)
## Summary

Building on the work in #220 (which adds UTC absolute datetime mode for the
reset timers), it would be very useful to support rendering the reset
timestamp in the user's local timezone — or any configurable timezone.

## Use case

I'm in `Asia/Tokyo` (UTC+9). Seeing `Resets 2026-04-25 10:00 UTC` requires me
to mentally add 9 hours to know when my block resets. A local-time format like
`Resets 7pm (Asia/Tokyo)` or `Resets 2026-04-25 19:00 JST` would be much more
actionable for daily planning.

This is also relevant to users who want to see reset time in:

- Their machine's local timezone (default behavior of most CLI tools)
- A specific timezone explicitly configured (for users in shared sessions or
  servers in different regions)

## Proposed behavior

After #220 lands, extend the date mode with timezone configuration:

```json
{
  "type": "block-reset-timer",
  "metadata": {
    "display": "date",
    "timezone": "local"
  }
}
```

Values for `timezone`:

- `local` (default): renders in the user's system timezone, formatted via
  `Intl.DateTimeFormat()`
- An IANA timezone name (e.g., `"Asia/Tokyo"`, `"America/New_York"`,
  `"Europe/Riga"`): explicit override
- `UTC` (current behavior in #220): preserved as the existing default if
  preferred for backwards compatibility

## Output examples

| timezone | output (compact=false) | output (compact=true) |
|---|---|---|
| `UTC` (current) | `2026-04-25 10:00 UTC` | `04-25 10:00Z` |
| `local` (Asia/Tokyo) | `2026-04-25 19:00 JST` | `04-25 19:00` |
| `America/New_York` | `2026-04-25 06:00 EDT` | `04-25 06:00` |

## Implementation notes

`Intl.DateTimeFormat()` with `timeZone` option handles this natively in
Node.js. Roughly:

```ts
new Intl.DateTimeFormat('en-CA', {
  timeZone: tz === 'local' ? undefined : tz,
  year: 'numeric',
  month: '2-digit',
  day: '2-digit',
  hour: '2-digit',
  minute: '2-digit',
  hour12: false,
  timeZoneName: 'short',
}).format(date);
```

## Related

- #220 (PR adding UTC date mode — this issue is a follow-up to that)
- #34 (closed: timezone configuration for block timer — similar theme)
- #139 (closed: block timer time zone bug)

isseeeeey55:
Closes #337.

Adds an optional `timezone` metadata key to `BlockResetTimer` and `WeeklyResetTimer`. When provided, the reset timestamp (already exposed in date mode via #220) is rendered through `Intl.DateTimeFormat` in the specified timezone instead of UTC.

## Motivation

Following #220, reset timers can show absolute timestamps — but only in UTC. For users outside UTC, mentally adding/subtracting offsets every time defeats the point of switching from countdowns to clock times. This change lets each timer widget render in any IANA timezone or the system local timezone.

## Behavior

| `metadata.timezone` value | Behavior |
|---|---|
| (omitted) | Preserves existing UTC output (no behavior change) |
| `\"UTC\"` | Same as omitted |
| `\"local\"` | Uses the system timezone via `Intl.DateTimeFormat()` |
| Any IANA TZ name (e.g. `\"Asia/Tokyo\"`) | Renders in that zone |
| Invalid TZ name | Falls back to UTC (so misconfiguration never blanks the widget) |

## Output examples

| timezone | output (compact=false) | output (compact=true) |
|---|---|---|
| (none) / `UTC` | `2026-03-12 08:30 UTC` | `03-12 08:30Z` |
| `Asia/Tokyo` | `2026-03-12 17:30 GMT+9` | `03-12 17:30` |
| `local` | `2026-03-12 17:30 <tz>` | `03-12 17:30` |

## Sample widget config

\`\`\`json
{
  \"id\": \"weekly-reset\",
  \"type\": \"weekly-reset-timer\",
  \"metadata\": {
    \"absolute\": \"true\",
    \"timezone\": \"Asia/Tokyo\"
  }
}
\`\`\`

## Implementation

- \`formatUsageResetAt\` grows an optional \`timezone\` parameter
- When non-UTC, uses \`Intl.DateTimeFormat\` with the \`timeZone\` option to assemble \`YYYY-MM-DD HH:MM <tzName>\` from \`formatToParts\`
- Both reset timer widgets read \`metadata.timezone\` via a new \`getUsageTimezone(item)\` helper and pass it through
- Invalid \`timeZone\` triggers \`Intl.DateTimeFormat\` to throw; we catch and fall back to the original UTC formatter

## Tests

New cases added to \`src/utils/__tests__/usage.test.ts\`:

- \`formats reset timestamps in a specific IANA timezone\` — Asia/Tokyo, both regular and compact
- \`keeps UTC behavior when timezone is undefined or \"UTC\"\` — backwards compat
- \`falls back to UTC when timezone is invalid\` — misconfiguration safety
- \`renders the system local timezone when timezone is \"local\"\` — pattern-based assertion to stay portable across CI environments

## Verification

\`\`\`bash
bun install
npx tsc --noEmit
npx vitest run src/utils/__tests__/usage.test.ts
npx vitest run src/widgets/__tests__/BlockResetTimer.test.ts src/widgets/__tests__/WeeklyResetTimer.test.ts
bun run build
\`\`\`

All pass locally.

## Notes

- No keybind / TUI configurator changes in this PR — \`timezone\` must currently be set by editing \`settings.json\` directly. A follow-up could add a TUI prompt to pick from common IANA names; happy to do that in a separate PR if desired.
- The output for \`local\` uses whatever short label \`Intl\` yields on the host (e.g. \`GMT+9\` on macOS V8 for Asia/Tokyo, since ICU returns the offset rather than \`JST\`). Improving the label substitution is out of scope here.

### DIFF
--- a/src/tui/components/items-editor/__tests__/input-handlers.test.ts
+++ b/src/tui/components/items-editor/__tests__/input-handlers.test.ts
@@ -641,6 +641,37 @@ describe('items-editor input handlers', () => {
         expect(customEditorState?.widget?.type).toBe('reset-timer');
     });
 
+    it('opens custom editor for reset timer locale action', () => {
+        const widgets: WidgetItem[] = [
+            { id: '1', type: 'reset-timer', metadata: { absolute: 'true' } }
+        ];
+        const onUpdate = vi.fn();
+        const setCustomEditorWidget = vi.fn();
+
+        handleNormalInputMode({
+            input: 'l',
+            key: {},
+            widgets,
+            selectedIndex: 0,
+            separatorChars: ['|', '-'],
+            onBack: vi.fn(),
+            onUpdate,
+            setSelectedIndex: vi.fn(),
+            setMoveMode: vi.fn(),
+            setShowClearConfirm: vi.fn(),
+            openWidgetPicker: vi.fn(),
+            getCustomKeybindsForWidget: (widgetImpl, widget) => widgetImpl.getCustomKeybinds ? widgetImpl.getCustomKeybinds(widget) : [],
+            setCustomEditorWidget
+        });
+
+        expect(onUpdate).not.toHaveBeenCalled();
+        const customEditorState = setCustomEditorWidget.mock.calls[0]?.[0] as
+            | { action?: string; widget?: WidgetItem }
+            | undefined;
+        expect(customEditorState?.action).toBe('edit-locale');
+        expect(customEditorState?.widget?.type).toBe('reset-timer');
+    });
+
     it('uses v to cycle skills widget mode', () => {
         const widgets: WidgetItem[] = [
             { id: '1', type: 'skills' }
--- a/src/tui/components/items-editor/__tests__/input-handlers.test.ts
+++ b/src/tui/components/items-editor/__tests__/input-handlers.test.ts
@@ -584,6 +584,63 @@ describe('items-editor input handlers', () => {
         expect(updated?.[0]?.metadata?.absolute).toBe('true');
     });
 
+    it('uses h to toggle reset timer hour format in timestamp mode', () => {
+        const widgets: WidgetItem[] = [
+            { id: '1', type: 'reset-timer', metadata: { absolute: 'true' } }
+        ];
+        const onUpdate = vi.fn();
+
+        handleNormalInputMode({
+            input: 'h',
+            key: {},
+            widgets,
+            selectedIndex: 0,
+            separatorChars: ['|', '-'],
+            onBack: vi.fn(),
+            onUpdate,
+            setSelectedIndex: vi.fn(),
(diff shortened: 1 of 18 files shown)


sirmalloc:
Thanks for this.  None of the widgets currently have settings only accessible via config, everything is surfaced in the TUI.  As such, I added a picker for the timezone and locale with fuzzy search and support for custom locales that aren't in the list.  I also added a 12/24hr time toggle.  I'll publish this shortly.

PR #344, opened 2026-04-27
fix(renderer): collapse separators around empty widgets

CorticalCode:
Spotted this while working on #282.

The status line renderer leaves an empty slot plus a leftover separator wherever a hide-capable widget renders null. The compaction-counter returns null when `count` is 0 and only shows ↻N after a compaction event — but the empty slot still occupied its layout position:

```
Model | branch | +5/-2 |  | Ctx: 42% | /path
```

**Scope**

Affects any widget that conditionally returns null:

- Git widgets with `hideNoGit: 'true'` (`git-branch`, `git-changes`, `git-staged-files`, etc.) in non-git contexts
- `git-conflicts` when there are no conflicts
- `git-pr` when no PR exists
- `compaction-counter` at count 0

Any layout using explicit `{type: "separator"}` items between such widgets is affected.

**Root cause**

The walkback in `renderStatusLine` that decides whether each separator should emit was looking for *any* prior widget with content, walking past empty ones. That correctly hides a leading separator (nothing before it) but leaves the orphan when an empty widget sits in the middle of a line.

**Fix**

Stop the walkback at the first non-separator widget and emit only when that widget actually rendered content.

- Auto-separator (`defaultSeparator`) path was already correct — it filters empty widgets out of its element chain. Not touched.
- Leading-edge behavior preserved.
- A `merge: 'no-padding'` chain that crosses a now-collapsed separator naturally glues to the next visible widget.

**Live-tested**

Layout: model, `git-branch` and `git-changes` (both `hideNoGit: 'true'`), `compaction-counter`, `context-percentage`, `current-working-dir`.

- Separators collapse cleanly when hide-capable widgets hide
- Separators remain when those widgets render content
- No regressions when all widgets are populated

**Tests**

9 new tests in `renderer-separator-collapse.test.ts`:

- Empty-in-the-middle case
- Consecutive empty widgets
- Leading empty (regression guard)
- Merge interaction
- `defaultSeparator` path independence
- All-content happy path

`bun run lint` and `bun test` both clean (955 tests).

### DIFF
--- a/src/utils/__tests__/renderer-separator-collapse.test.ts
+++ b/src/utils/__tests__/renderer-separator-collapse.test.ts
@@ -1,0 +1,150 @@
+import {
+    describe,
+    expect,
+    it
+} from 'vitest';
+
+import type { RenderContext } from '../../types/RenderContext';
+import {
+    DEFAULT_SETTINGS,
+    type Settings
+} from '../../types/Settings';
+import type { WidgetItem } from '../../types/Widget';
+import { renderStatusLine } from '../renderer';
+
+interface PreRenderedWidget {
+    content: string;
+    plainLength: number;
+    widget: WidgetItem;
+}
+
+function createSettings(overrides: Partial<Settings> = {}): Settings {
+    return {
+        ...DEFAULT_SETTINGS,
+        ...overrides,
+        powerline: {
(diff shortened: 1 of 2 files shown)


sirmalloc:
Thanks, I'll publish in the next release.

Issue #259, opened 2026-03-23
Unified editing: merge color editing into Edit Lines via Tab mode


hangie:
## Summary

Major TUI overhaul that unifies the editing experience with consistent navigation patterns throughout. What started as a proposal to merge color editing into Edit Lines grew into a comprehensive rethink of TUI navigation.

## Problem (original)

Editing a widget's color required navigating: **MainMenu → Edit Colors → LineSelector → ColorMenu → (edit) → ESC × 3 → MainMenu**. The "Edit Lines" and "Edit Colors" top-level split put the same objects behind different doors.

Additionally, the Rules Editor was a full-screen overlay that removed all context of the widget list. Navigation was inconsistent across screens — some wrapped, some didn't; some used ← Back menu items, some used ESC only; keybinds were case-sensitive in some screens and case-insensitive in others.

### 1. Unified Color Editing (Tab mode)
- Press **Tab** to switch between items mode and color mode on any widget
- Inline color editing: ←→ cycle, hex, ANSI256, bold, fg/bg toggle
- Removed separate "Edit Colors" menu entirely
- Deleted `ColorMenu.tsx` and `color-menu/mutations.ts`

### 2. Accordion Rules Editor
- Press **→** on a widget to expand its rules inline below it
- Rules use the **same interaction patterns** as widgets: ↑↓ navigate, Tab for color/property mode, Enter for focus mode
- Deleted full-screen `RulesEditor.tsx` overlay — replaced by accordion
- Empty rules show placeholder: `(no rules — press 'a' to add)`

### 3. Consistent Tree Navigation
- **→ drills in, ← drills out** at every level: LineSelector → ItemsEditor → Rules → ConditionEditor
- **Enter = focus/edit mode** at every level: ↑↓ reorders, ←→ edits properties
- **ESC peels back one layer** — never skips levels
- **↑↓ wraps** at top/bottom everywhere
- Removed all "← Back" menu items — ← arrow handles back navigation

### 4. Polish
- Exit confirmation when unsaved changes exist
- Case-insensitive keybinds across all screens
- "Enter to edit" replaces "Enter to move" for better discoverability
- PowerlineSeparatorEditor refactored to same Enter/focus pattern

## Navigation Model

| Level | ↑↓ | → | ← | Enter | Tab |
|-------|-----|---|---|-------|-----|
| MainMenu | Navigate | Select | — | Select | — |
| LineSelector | Navigate | Enter editor | Back | Move mode | — |
[...]

PR #346, opened 2026-04-29
feat: color editing as a widget-editor mode, with per-widget overrides under Powerline themes

hangie:
## Summary

This PR started as Tab navigation between the item and color editors. Following @sirmalloc's five points in the thread below, it now also implements per-widget color overrides under Powerline themes and reshapes color editing into a mode of the widget editor.

Related to #259.

## The five points

**1. Allow editing colors when a theme is applied; save modified colors with their widgets.**

A widget can **pin** its foreground and/or background, independently. A pinned channel keeps the widget's own color and overrides the theme; an unpinned channel takes the theme. Colors are stored on the widget as they always were — the pin is two optional booleans (`pinColor`, `pinBackgroundColor`), so there is no new color storage and no settings migration.

The guard that blocked color editing under a managed theme is gone.

**2. Main menu color entry becomes a signpost.**

`🎨 Edit Colors` no longer opens the standalone color flow. It lands on a notice explaining that color editing lives in the widget editor, with an action that jumps straight to the line selector. The entry is left in place for you to remove when you're ready.

The old color line-selector screen became unreachable and was deleted, which also removed the "colors are managed by the Powerline theme, press any key to go back" block — the wall this PR exists to replace.

**3. Tab to color editing, keeping the widget editor's shape; Esc back one menu.**

Both editors now render their rows through one shared component, so a row is *the same row* in both modes: same numbering, same indicator, same structure markers (`(merged→)`, `(raw value)`, `(no-align)`), tinted with the color the widget actually renders in. Titles share a stem and differ by a mode tag — `Edit Line 1 [WIDGETS]` / `Edit Line 1 [COLORS] [FOREGROUND]`. A parity test asserts the rows are byte-identical between modes, ANSI included.

Esc backs out to the line selector from either mode, and the same one.

`ink-select-input` is gone from the color editor (it made a shared row impossible), which also removed a remount-on-highlight hack and a duplicate static list. Nothing else imported it, so it's dropped from `package.json`.

**4. Color editing stays familiar; the rendering path honours per-widget overrides.**

`renderPowerlineStatusLine` applies a pinned channel over the theme. `preserveColors` still wins for the foreground of a custom command. The theme color index still advances for pinned widgets, so pinning one widget doesn't shift its siblings' colors.

The editor now shows what actually renders: an unpinned channel displays the theme's color and reports it as `(theme)`, rather than the dormant stored value it used to show. Both the row tinting and the status line read from one helper that derives colors from the same slot assignment the renderer uses, so the preview cannot drift from the output.

**5. Theme switcher prompts to keep or remove overrides.**

Changing the theme with pins present prompts to keep or clear them.

## The hard requirement

> Launching an existing statusline config with a given powerline theme will not change appearance when opened in the updated app.

Pins are opt-in and optional, so an existing config has none: every channel takes the theme exactly as before, and nothing is rewritten on load. The tricky case you named — colors persisted *before* a theme was enabled — stays dormant, because only a conscious pin makes a color win.

`src/utils/__tests__/renderer-theme-color-override.test.ts` codifies this, including the pre-existing-color case.

## UX decisions worth a look

- **A channel must be pinned before its color can be edited.** Editing an unpinned channel used to pin it automatically, which made a stray `←`/`→` destructive: it pinned *and* overwrote the dormant color, and since unpinning deliberately keeps the color, the original was unrecoverable. Under a theme the color keys now do nothing until you press `p`, and the status line says `- theme applies, press (p) to override` so they never look broken. Bold, dim, reset and clear-all aren't theme-driven and stay live. Nothing changes when no theme is active.
- **Pinning takes over what you're looking at** — your own color if the widget has one, otherwise the theme color it is currently rendering. Taking control never changes the appearance by itself.
- **Unpinning is non-destructive**: it hands the channel back to the theme and keeps your color, so re-pinning restores it.
- **Pin state is shown per row** (`(fg pinned)`, `(bg pinned)`, `(fg+bg pinned)`) in both modes, so every override is visible at a glance rather than only for the highlighted widget.
- **Row parity was deliberately not attempted.** The color list still shows only colorable widgets, so rows are numbered by position in the line and a skipped separator leaves a gap. Tab only enters color editing from a colorable widget, so you can never land on a row that rejects input; padding the list with unselectable rows seemed worse than the gap. Now that `ink-select-input` is gone, full parity is a small change if you'd prefer it.
- **The last commit is separable.** `feat(items-editor): tint widget rows with the colour they render in` makes the widget editor adopt color-mode rendering, which costs its green selected-row highlight (selection is the `▶` indicator, as in color mode; move mode keeps its blue row). That inverts the direction of your request, so it sits last and can be dropped on its own without touching anything else.

## Also included

- One theme-slot assignment now serves both the renderer and the editor, so "which theme color does this widget get" is decided in one place.
- Fixed a pre-existing flake in `PowerlineThemeSelector.test.ts`: two ink-driven tests advanced on fixed 25 ms sleeps and failed under load. They now poll for the state each step waits on. (Ink writes a frame *before* the next screen's input handler attaches, so a keypress sent the instant a frame appears is dropped — worth knowing for other TUI tests.)
- `docs/USAGE.md` documents the color editing mode, the pin model, and the keybinds.

## Not included, deliberately

Enabling Powerline permanently deletes manual separators and disabling doesn't restore them. That's pre-existing and yours to decide on, so it's raised separately as #531 rather than changed here.

## Test plan

- [x] `bun run lint` clean (tsc + eslint, no warnings)
- [x] `bun test` — 1715 pass, 0 fail
- [x] `bun run build` succeeds
- [x] Regression test for the no-appearance-change guarantee, including pre-theme colors
- [x] Renderer tests for pinned fg/bg, `preserveColors`, and index advancement across pinned and merged widgets
- [x] Editor tests for the pin gate, pin seeding, theme-color display, row parity, and the mode tags
- [x] Manual TUI testing


sirmalloc:
Running into some issues with this one.  If you have a powerline managed theme turned on, go to the widget editor and press Tab, it takes you in to edit the colors (which should not be allowed when a managed theme is active).  Changing the colors has no effect (as expected when a theme is active).  Pressing ESC while editing the colors takes you to the screen with the message "⚠ Colors are currently managed by the Powerline theme: Minimal", and pressing any key takes you back to the main menu with the 'Edit Colors' item highlighted.

Throughout the TUI, the Esc key always takes you back to the previous screen, but pressing Tab to go to the color editor breaks that UX.  I feel, at a bare minimum, going into color editing mode via Tab should take you out of color editing mode and back to the widget editor when either Tab or Esc is pressed to maintain the user's navigation context.

If a powerline theme is active, an appropriate solution might be to prompt the user to customize the theme (using the same 'c' key functionality when on the theme list), while warning them that the theme's colors for each widget will be copied and then they can customize them.

I'm not sure I fully like this flow though, as customizing a theme disables auto-applying the theme to newly added widgets.  Everything added after customizing a theme gets the default, unthemed colors.  So by presenting the Tab as a shortcut to easily customize widget colors, then prompting them to customize it at that point in the flow might confuse users why the theme they started with is no longer being applied.

I'm open to suggestions on this, I like the quick access but it doesn't work well when a theme is active.

hangie:
> If you have a powerline managed theme turned on, go to the widget editor and press Tab, it takes you in to edit the colors (which should not be allowed when a managed theme is active).

Fixed in cbcd811 — when a managed theme is active, onTabSwap is no longer passed to either editor, suppressing both the hint and the keypress. This is consistent with the existing blockIfPowerlineActive guard on the canonical color editing path.

If a user wants per-widget color control, they can already customize the theme via Powerline Settings → Theme Selector → c, which bakes the theme colors into widget fields and switches to custom mode — Tab swap then becomes available.

Did you want a different behaviour, or is this consistent with your intent?

hangie:
> I feel, at a bare minimum, going into color editing mode via Tab should take you out of color editing mode and back to the widget editor when either Tab or Esc is pressed to maintain the user's navigation context.

Tab already swaps back, so that works. For ESC, the current behaviour was deliberate — Tab is treated as a lateral jump between parallel editors, not a navigation that creates a back-stack. So ESC follows the canonical path for whichever editor you're in (ColorMenu → colorLines → main).

The alternative would be to track that you arrived via Tab and have ESC return you to the originating editor. If we go that route, it would be a single breadcrumb — repeated Tab presses just toggle between editors, they don't accumulate, so you'd never need to ESC multiple times to unwind.

Happy to implement either approach — do you have a preference?

hangie:
@sirmalloc - did you want me to implement the breadcrumbs as described? Or do you have other ideas?

sirmalloc:
> @sirmalloc - did you want me to implement the breadcrumbs as described? Or do you have other ideas?

@hangie Hang tight for a couple days on this one.  I wanted to modify the theme rendering pipeline first to allow keeping a theme on and overriding colors for individual widgets, then I was considering removing the Color editor entirely from the main menu in favor of making the widget editor the sole entry point into color customization.  I think it flows better and if we can apply colors while themes are on it'll feel pretty natural no matter what mode the user is in.

I'll probably keep the main menu item around for a couple weeks and direct people where to edit colors now so existing users aren't confused.  Thanks for your patience on this one.

sirmalloc:
Sorry I've been slow on this. Still on my radar, making time to get through all these PRs soon.

hangie:
@sirmalloc — checking in on this one. It's been a few weeks, so no rush, but wanted to see where your thinking landed.

I've kept the branch current with main (merges clean, tests green). From what I can see, the theme-rendering changes — keeping a theme on while overriding per-widget colors — haven't landed yet, so the current cbcd811 guard (disabling Tab-swap under a managed theme) still holds for today's behaviour.

Happy to treat this PR as a building block for the broader change you described, or to adjust it to fit whatever shape you land on. Do you want me to proceed with the breadcrumb ESC behaviour now, hold for your pipeline work, or rethink the entry point?

hangie:
@sirmalloc - any updates?
I think you could take this as your base for the editor approach and subsequent removal of the color editor from the top level menu. Reach out to discuss if you like. My other PR for the rules editor depends on this, and if you intend to take a different path instead of take this PR, then I'll have to abandon that or rewrite it on top of whatever changes you do differently for navigation and editors...

sirmalloc:
> @sirmalloc - any updates? I think you could take this as your base for the editor approach and subsequent removal of the color editor from the top level menu. Reach out to discuss if you like. My other PR for the rules editor depends on this, and if you intend to take a different path instead of take this PR, then I'll have to abandon that or rewrite it on top of whatever changes you do differently for navigation and editors...

@hangie I've been a bit busy lately with some career changes so I haven't been able to give the project proper attention aside from merging some low hanging fruit.  Here's where I'm at, and if you want to try to update this PR to accommodate that, I'll be happy to merge it:

1. Modify color logic to allow editing colors when a theme is applied.  Save the modified colors only with their respective widgets.  There might need to be some UX work around the powerline theme screen and the whole '(c)ustomize theme' concept.
2. Make the main menu 'color editing' screen show users a message that the functionality has been moved to the widget editing screen, and offer an option to go there.  We'll eventually remove this.
3. Do the TAB to edit colors on the widget editing screen.  Ideally the UI will keep most of it's 'shape' from the widget editor, but it'll now be in 'Color Editing mode', and allow pressing TAB to switch back.  Esc will just go back one menu to the line selector.
4. Color editing will stay mostly the same UX as it is now, and the rendering path will be fixed to override the powerline colors for a given widget's foreground / background.
5. Powerline theme switcher should have a prompt when the theme is changed to keep or remove custom color overrides.

One hard requirement that should be codified into regression tests:

1. Launching an existing statusline config with a given powerline theme will not change appearance when opened in the updated app.  You have to be careful here because if the user had colors configured for a widget before enabling a powerline theme, they might still be persisted in the config but ignored during rendering.  We need a solution so when the update drops, nobody has their colors changed automatically and think that things broke.

Again, sorry for dropping the ball on this one.  Let me know if you want to discuss further, I'm open to suggestions.

hangie:
@sirmalloc — no worries at all on the timing, and thanks for laying out where you've landed. I'm glad to take this on and see it through. Your five points all make sense and I'm on board, including the details: ESC dropping back to the line selector, colour editing becoming a "Colour Editing mode" that keeps the widget editor's shape with TAB to switch back, the old top-level colour menu turning into a signpost, and the theme switcher prompting to keep/remove overrides when the theme changes. My managed-theme guard comes out as part of this — there'll be real behaviour behind it now.

Here's the shape I intend for the colour model, since your hard requirement (no appearance change for existing themed configs) is what drives it:

- **Per-widget overrides via a per-channel pin.** A widget can pin its foreground and/or background independently so its own colour wins over the theme. The renderer gains that override step; anything unpinned still takes the theme.
- **Nothing changes on upgrade — and no migration.** Existing widget colours stay dormant under a theme exactly as today; only a *conscious pin* makes a colour win. So opening an existing themed config renders identically, with no settings rewrite. I'll codify that as regression tests up front, per your hard requirement — including the tricky case of colours that were persisted before a theme was enabled.
- **Non-destructive editing.** Pinning a widget surfaces your existing colour (or the current theme colour if you have none) as the starting point, so you edit from what's there rather than from scratch. Unpinning just hands the channel back to the theme — it never wipes your colour.
- **One colour per widget.** A widget's colour is the same with or without a theme; the pin only decides whether the theme may override it. (Genuinely *different* colours with vs without a theme would be a clean additive later if it's ever wanted — but I don't think it is, and one colour matches your "override the theme per widget" framing.)

Your keep/remove prompt on a theme change (point 5) then maps cleanly onto clearing pins.

I'll follow up with a short written plan — data model, render precedence, the editing flow, and the regression-test shape — so you can gut-check it before I go deep. Happy to hop on a thread if that's easier around your schedule. Thanks again for getting this unblocked.

sirmalloc:
@hangie  Sounds good on all your points.  I'd be happy to review either a full plan or just test the implementation, either way.

hangie:
@sirmalloc - let me know what you think. Implementation complete.

PR #349, opened 2026-05-01
fix: default to full flex mode and make multi-line alignment independent per line

godlockin:
## Summary

- **Default flexMode changed from `full-minus-40` to `full`**: New/default configurations were unnecessarily truncating status line content by subtracting 40 characters from terminal width. Changed to `full` (subtracts only 6) so content is visible by default. Existing configs with an explicit `flexMode` value are unaffected.

- **Per-line independent column widths for autoAlign**: `calculateMaxWidthsFromPreRendered` previously computed a single shared `number[]` of max column widths across *all* lines. This caused a wide widget on one line to force the same column wider on every other line. Changed return type to `number[][]` so each line gets its own independent column widths.

## Test plan

- [x] All 1147 existing tests pass (`bun test`)
- [x] TypeScript type checks clean (`npx tsc --noEmit`)
- [x] Renderer-specific tests (`renderer-ansi`, `renderer-flex-width`, `renderer-powerline-theme`, `renderer-separator-collapse`) all pass


sirmalloc:
@godlockin Can you explain the intent of this PR a little more clearly? The whole purpose of auto-alignment in powerline mode is to ensure every widget is vertically aligned with the widget above and/or below it. By calculating max width's on a per-line basis, this effectively prevents that functionality from working as intended.

A better fix to exclude wide widgets from affecting everything else would be to allow a single widget to be manually flagged as excluded from the auto-align calculation, which would basically cause that widget and everything after it on the same line to be exempt from any changes to their default widths.

godlockin:
Sorry replay late, I didn't get the notice of your comment.

This change is aim to make the status' display dynamic adaption the width of terminal window

When I change the width of terminal window (and of cuz split it .etc) the cc status may have some display issue like omitted a part, or even can't display while there is still some blank space

PR #350, opened 2026-05-01
Modify the display in multi-line

godlockin:
Reported behavior: on multi-line status bars, a widget with long content (e.g. a long git branch name) would push adjacent columns wider on other lines. Also, out-of-the-box configurations were truncating content on wider terminals due to an overly conservative default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,16 @@ npx -y ccstatusline@latest
 bunx -y ccstatusline@latest
 ```
 
+### Build from source (for custom/forked versions):
+
+```bash
+git clone git@github.com:godlockin/ccstatusline.git
+cd ccstatusline
+./install.sh
+```
+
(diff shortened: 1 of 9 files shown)


sirmalloc:
@godlockin Can you explain the intent of this PR a little more clearly?  The whole purpose of auto-alignment in powerline mode is to ensure every widget is vertically aligned with the widget above and/or below it.  By calculating max width's on a per-line basis, this effectively prevents that functionality from working as intended.

A better fix to exclude wide widgets from affecting everything else would be to allow a single widget to be manually flagged as excluded from the auto-align calculation, which would basically cause that widget and everything after it on the same line to be exempt from any changes to their default widths.

sirmalloc:
Closing this as a dupe of #349, I'll copy my previous question there.

PR #370, opened 2026-05-13
fix: ignore transient 0% flashes in compaction counter

KorenKrita:
## Problem

The compaction counter produces false counts when Claude Code emits incomplete status JSON frames that briefly report `used_percentage` as 0%. Each 0% flash resets `prevCtxPct` to 0, and the next valid reading triggers a large drop detection — incrementing the compaction count even though no compaction occurred.

## Fix

Ignore `currentCtxPct` values below 1% in `detectCompaction()`. Real compaction never drops the percentage that low, so any sub-1% reading is a transient glitch from an incomplete status JSON frame.

Changes:
- Added `MIN_CTX_PCT = 1` constant
- Changed guard from `currentCtxPct < 0` to `currentCtxPct < MIN_CTX_PCT`
- Updated the "starts at 0%" test to use 1% (0% is now filtered)
- Added tests for 0% and sub-1% glitch filtering

## Testing

All 1306 tests pass, lint clean.


sirmalloc:

PR #377, opened 2026-05-13
fix(terminal): use stty -F/-f to read tty size in detached spawn

### LINKED ISSUE #376: Terminal width probe fails under Claude Code ≥ 2.1.139 (no controlling TTY for spawned statusline)
## Summary

On Linux under Claude Code ≥ 2.1.139, ccstatusline's TTY-width probe fails entirely and falls back to `tput cols → 80`. With the default `flexMode: "full-minus-40"`, the statusline truncates at 40 columns regardless of actual terminal width.

## Root cause

[Claude Code 2.1.139 changelog](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#21139):

> Fixed a bug where a hook writing to the terminal could corrupt an on-screen interactive prompt; **hooks now run without terminal access.**

The statusLine spawn is hardened the same way. The ancestor walk in `probeTerminalWidth` correctly finds the controlling PTY, but `getWidthForTTY`'s `stty size < /dev/${tty}` form fails with `ENOTTY` when the calling process has no controlling terminal.

## Reproduction

```
$ stty size < /dev/pts/0
stty: Inappropriate ioctl for device
$ stty -F /dev/pts/0 size
36 159
```

Reproducible without ccstatusline via `setsid -w bash -c 'stty size < /dev/pts/0'` — confirms it's generic, not terminal-emulator-specific.

## Fix

Use `stty -F <path> size` (GNU coreutils) or `stty -f <path> size` (BSD); they open the device internally and succeed regardless of controlling-tty status. PR coming.

## Related

Open requests to expose `terminal.width` in the statusLine JSON, which would future-proof this independently:

- anthropics/claude-code#5430
- anthropics/claude-code#22115

AgarwalPragy:
## Summary

Fixes the TTY-width probe under Claude Code ≥ 2.1.139, which [spawns the statusline without a controlling terminal](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#21139):

> Fixed a bug where a hook writing to the terminal could corrupt an on-screen interactive prompt; **hooks now run without terminal access.**

The existing `stty size < /dev/${tty}` form fails with `ENOTTY` in that context.

Closes #376.

## Change

`src/utils/terminal.ts` — `getWidthForTTY` now tries three variants in order; first positive integer wins:

1. `stty -F <path> size` (GNU coreutils, Linux)
2. `stty -f <path> size` (BSD, macOS)
3. `stty size < <path>` (legacy fallback)

## Test plan

- New unit test for the BSD fallback case (`-F` errors, `-f` succeeds)
- Existing probe tests updated for the new command shape
- All 1303 tests pass; lint clean
- Live-verified on Linux + Claude Code 2.1.140: detection goes from null → real terminal width

### DIFF
--- a/src/utils/__tests__/terminal.test.ts
+++ b/src/utils/__tests__/terminal.test.ts
@@ -46,7 +46,7 @@ describe('terminal utils', () => {
                 return 'ttys001\n';
             }
 
-            if (command === `stty size < /dev/ttys001 | awk '{print $2}'`) {
+            if (command === `stty -F /dev/ttys001 size 2>/dev/null | awk '{print $2}'`) {
                 return '120\n';
             }
 
@@ -57,7 +57,7 @@ describe('terminal utils', () => {
         expect(mockExecSync.mock.calls.map(([command]) => command)).toEqual([
             `ps -o ppid= -p ${process.pid}`,
             'ps -o tty= -p 1234',
-            `stty size < /dev/ttys001 | awk '{print $2}'`
+            `stty -F /dev/ttys001 size 2>/dev/null | awk '{print $2}'`
         ]);
     });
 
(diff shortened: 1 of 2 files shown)


sirmalloc:
Thanks, I'll publish this in the next release

Issue #308, opened 2026-04-15
Pass terminalWidth in custom command stdin JSON


LZong-tw:
## Summary

Custom Command widgets receive `context.data` as JSON via stdin, but `context.terminalWidth` is not included. This makes it impossible for scripts to adapt output based on terminal width.

## Current behavior

From the source (`dist/ccstatusline.js:57125`):
```js
const jsonInput = JSON.stringify(context.data);
let output = execSync3(item.commandPath, {
  encoding: "utf8",
  input: jsonInput,
  // ...
});
```

`context.terminalWidth` exists but is not part of `context.data`, so custom commands can't access it. Standard detection methods (`tput cols`, `stty size`, `$COLUMNS`) all return 80 in the piped stdio context.

## Proposed change

Include `terminalWidth` (or `terminal_width`) in the JSON passed to custom commands:

```json
{
  "workspace": { "current_dir": "/path/..." },
  "model": { "display_name": "..." },
  "terminal_width": 142
}
```

## Use case

Adaptive label display — show full labels ("ReadCache", "CacheCreate") on wide terminals and abbreviated ("RC", "CC") on narrow ones.

## Workaround

Currently using full labels and relying on ccstatusline's own truncation ("..."), which is suboptimal on narrow screens.

PR #380, opened 2026-05-14
feat(terminal): honor `CCSTATUSLINE_WIDTH` env var to override probe

### LINKED ISSUE #376: Terminal width probe fails under Claude Code ≥ 2.1.139 (no controlling TTY for spawned statusline)

danielnc:
## Summary

Adds an explicit width override so users can bypass the TTY probe entirely when both ancestor-walk and `tput cols` fall through. Set `CCSTATUSLINE_WIDTH=<cols>` on the statusLine command, get exactly that width.

## Motivation

PR #377 (`stty -F`/`stty -f`) closed #376 by fixing the case where an ancestor process holds a TTY but the legacy `stty size < /dev/tty` form errors with `ENOTTY` (Claude Code ≥ 2.1.139 spawns statusline/hooks without terminal access). That fix lands the common path.

It does not cover the case where **no ancestor process owns a TTY at all** — observed under some Claude Code agent-mode spawn paths, IDE integrations (Cursor, JetBrains plugin), and nested-shell scenarios. The ancestor walk runs out of TTY-bearing PIDs, the `tput cols` fallback returns 80, and `flexMode: "full-minus-40"` truncates the multi-line layout to **40 columns** regardless of the real terminal width.

Concretely: on CC 2.1.141 + ccstatusline 2.2.17 + a wide iTerm2, the `terminal-width` widget reports `Term: 80` in this configuration. The fallback to `tput cols` is the symptom; the absence of any TTY ancestor is the root cause.

This patch is complementary to #377 — it adds an explicit user-controlled escape hatch for cases the probe cannot solve on its own, and gives users a knob today while upstream work on passing `terminalWidth` via stdin JSON (#308) lands.

## Change

`src/utils/terminal.ts` — `probeTerminalWidth` reads `CCSTATUSLINE_WIDTH` before any platform check or probe call. A valid positive integer short-circuits with that value. Anything else (missing, empty, non-numeric, zero, negative) falls through to existing logic — no behavior change for users who don't set it.

```ts
const overrideRaw = process.env.CCSTATUSLINE_WIDTH;
if (overrideRaw) {
    const override = parsePositiveInteger(overrideRaw);
    if (override !== null) {
        return override;
    }
}
```

## Usage

In `~/.claude/settings.json`:

```json
"statusLine": {
  "type": "command",
  "command": "CCSTATUSLINE_WIDTH=200 ccstatusline"
}
```

(or via a wrapper script, or any other mechanism that sets the env on the spawned process.)

## Test plan

Four new cases in `src/utils/__tests__/terminal.test.ts`:

- override short-circuits probing entirely (asserts `execSync` is never called)
- non-positive override (`"0"`) falls back to probing → returns probed width
- non-numeric override (`"wide"`) falls back to probing → returns probed width
- override applies on Windows where probing is otherwise disabled (asserts `canDetectTerminalWidth()` returns `true` and `execSync` is never called)

```
$ bunx vitest run src/utils/__tests__/terminal.test.ts
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Full-suite delta vs `main`: **+4 passing tests, 0 new failures**. The 105 pre-existing failures in `claude-settings.test.ts`, `compaction.test.ts`, `global-command-resolution.test.ts`, `ClaudeAccountEmail.test.ts` reproduce identically on `main` and on this branch — they're env-specific (platform/shell paths) and unrelated to terminal probing.

```
$ bun run lint
$ # clean, exit 0
```

## Related

- Closes the remaining "no ancestor TTY" gap left by #377
- Workaround on the user side today without requiring `terminalWidth` in stdin JSON (#308)
- Cross-references: anthropics/claude-code#5430, anthropics/claude-code#22115

## Risk

- Behavior change is opt-in via env var; users who don't set `CCSTATUSLINE_WIDTH` get identical probe behavior to today.
- Invalid env values (empty, non-numeric, ≤ 0) fall through — same probe path as no env at all.
- Cross-platform: the override happens before the Windows short-circuit, so Windows users gain a working width source for the first time when they set it (previously `null`).


sirmalloc:

PR #381, opened 2026-05-15
fix(git): always pass --no-optional-locks to prevent index.lock races

BastianTrifork:
## Problem

When the statusline refreshes while the user has a concurrent git op in flight (e.g. `git restore --staged ...`), commands like `git diff --shortstat` race on `.git/index.lock` and the user's command fails with:

```
fatal: Unable to create '/path/.git/index.lock': File exists.
```

`git diff` takes the index lock to refresh stat cache entries; under the default `core.optionalLocks=true` that lock is contended even though diff is read-only from the user's perspective.

## Fix

`runGitArgs` now prepends `--no-optional-locks` to every git invocation, so the statusline can never block or interfere with the user's working git operations. The flag was already applied per-call to `status` (`getGitStatus`, `getGitFileStatusCounts`); those manual prefixes are now redundant and removed.

## Tests

- Arg-order assertions in `git.test.ts`, `git-remote.test.ts`, and four widget tests updated to expect the new prefix.
- `bun test`: 1333 / 1333 pass.
- `bun run lint`: clean.

## Notes

- `--no-optional-locks` is a top-level git option (git 2.15+); no behavioural change for read-only commands beyond skipping the stat-cache refresh.
- Caught locally by a watchdog poll on `.git/index.lock` — culprit identified as `getGitChangeCounts` → `git diff --shortstat`.

### DIFF
--- a/src/utils/__tests__/git-remote.test.ts
+++ b/src/utils/__tests__/git-remote.test.ts
@@ -229,7 +229,7 @@ describe('git-remote utils', () => {
             getRemoteInfo(remoteName, {});
 
             expect(mockExecFileSync.mock.calls[0]?.[0]).toBe('git');
-            expect(mockExecFileSync.mock.calls[0]?.[1]).toEqual(['remote', 'get-url', '--', remoteName]);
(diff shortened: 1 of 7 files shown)


sirmalloc:
Thanks, I'll get this published shortly.

PR #388, opened 2026-05-18
fix(extra-usage): correct unit assumptions for usage API response

### LINKED ISSUE #387: Extra usage widgets use wrong unit assumptions for usage API response
## Summary

The `ExtraUsageRemaining` and `ExtraUsageUtilization` widgets make incorrect unit assumptions about fields returned by the usage API, producing wrong displayed values for pay-as-you-go users.

## Details

In `src/widgets/ExtraUsageRemaining.ts`:
- The code assumes `extraUsageLimit` is in cents but `extraUsageUsed` is in dollars.
- In practice, **both fields are in cents**. The current math subtracts a cents value from a dollars value, producing a wildly inflated "remaining" amount.

In `src/widgets/ExtraUsageUtilization.ts`:
- The code multiplies `extraUsageUtilization` by 100, assuming it's a fraction (0–1).
- In practice, `extraUsageUtilization` is **already a percentage (0–100)**. Multiplying by 100 produces values that get clamped to 100% almost immediately.

## Expected behavior

- "Overage Left" should reflect the true remaining overage budget in dollars.
- "Overage Utilization" should reflect the actual percentage reported by the API.

## Fix

PR incoming that divides `extraUsageUsed` by 100 before subtracting, and removes the `* 100` on `extraUsageUtilization`.

roribio:
## Summary

Fixes #387.

- `ExtraUsageRemaining`: `extraUsageUsed` is in cents (not dollars), so divide by 100 before subtracting from the limit. Previously the widget subtracted a cents value from a dollars value, producing an inflated "Overage Left" amount.
- `ExtraUsageUtilization`: `extraUsageUtilization` is already a percentage (0–100), not a fraction. Removed the erroneous `* 100` that caused the value to clamp to 100% almost immediately.

## Test plan

- [x] Verify "Overage Left" shows a sane dollar value against a real pay-as-you-go account
- [x] Verify "Overage Utilization" matches the percentage reported by the Claude usage API
- [x] `bun run lint` passes
- [ ] `bun test` passes

### DIFF
--- a/src/utils/__tests__/usage-fetch.test.ts
+++ b/src/utils/__tests__/usage-fetch.test.ts
@@ -308,8 +308,8 @@ describe('fetchUsageData error handling', () => {
         extra_usage: {
             is_enabled: true,
             monthly_limit: 400000,
-            used_credits: 106,
-            utilization: 0.026
+            used_credits: 10600,
+            utilization: 2.6
         }
     });
     const rateLimitedResponseBody = JSON.stringify({
@@ -571,8 +571,8 @@ describe('fetchUsageData error handling', () => {
                 weeklyResetAt: '2030-01-07T00:00:00.000Z',
                 extraUsageEnabled: true,
                 extraUsageLimit: 400000,
-                extraUsageUsed: 106,
-                extraUsageUtilization: 0.026
+                extraUsageUsed: 10600,
+                extraUsageUtilization: 2.6
             });
             expect(result.second).toEqual(result.first);
             expect(result.requestCount).toBe(1);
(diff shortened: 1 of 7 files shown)


sirmalloc:
Thanks for the contribution here. The live API details you included were helpful.

I pushed a small follow-up commit with minor test/contract tweaks so the source-layer fixtures and `UsageData` comments match the units described in this PR:

  - `extra_usage.used_credits` is represented as cents
  - `extra_usage.utilization` is represented as a 0-100 percentage

The widget changes look good with that contract clarified. This will be included in the next release.

PR #396, opened 2026-05-22
feat(custom-command): include terminal_width in stdin JSON

### LINKED ISSUE #308: Pass terminalWidth in custom command stdin JSON

CorticalCode:
## Summary

Custom Command widgets receive `context.data` as JSON over stdin, but the terminal width was not part of it, so scripts had no way to adapt their output to the terminal. `tput cols` / `stty size` / `$COLUMNS` all report `80` in ccstatusline's piped stdio context, so the width is otherwise unrecoverable from a custom command. Requested in #308.

## What changed

- **`src/ccstatusline.ts`** — populate `terminalWidth` once on the render context via `getTerminalWidth()`. Custom commands execute during `preRenderAllWidgets`, *before* `renderStatusLine` computes the width, so the value has to be on the context up front. As a side benefit, the renderer now reuses this value instead of re-probing per line.
- **`src/widgets/CustomCommand.tsx`** — add `terminal_width` to the JSON piped to the command when a numeric width is known (omitted otherwise).

```json
{ "model": { "display_name": "..." }, "workspace": { "current_dir": "..." }, "terminal_width": 142 }
```

## Notes

- **Field name** `terminal_width` (snake_case) matches the existing Claude Code payload (`session_id`, `current_dir`, `used_percentage`, …).
- **Raw terminal width**, not flex-adjusted — scripts get the actual column count (and it honors `CCSTATUSLINE_WIDTH`), independent of `flexMode`. This is complementary to #380: `CCSTATUSLINE_WIDTH` overrides the probe, whereas this exposes the resulting width to custom-command stdin.
- **Omitted when undetectable**, so script authors can guard for its absence.

## Use case

Adaptive labels — full text ("ReadCache", "CacheCreate") on wide terminals, abbreviated ("RC", "CC") on narrow ones — as described in the issue.

## Testing

- New unit test (`src/widgets/__tests__/CustomCommand.test.ts`) mocking the `execSync` boundary: field present when width is known, existing data still passes through, field omitted when width is unknown.
- Full suite green (1396 tests), lint clean.
- Manually verified end-to-end with an adaptive custom-command script through both `bun` and the bundled `dist` (via `node`): full labels on wide terminals, abbreviated on narrow, the real tty width picked up with no override, and the raw width preserved under `full-minus-40`.
- `docs/USAGE.md` updated.

Closes #308

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -203,6 +203,7 @@ Add a single symbol or emoji to your status line when you want a compact visual 
 Execute shell commands and display their output dynamically:
 - Refreshes whenever the statusline is updated by Claude Code
 - Receives the full Claude Code JSON data via stdin (model info, session ID, transcript path, etc.)
+- Also includes `terminal_width` — the detected terminal width in columns, added by ccstatusline (omitted when it can't be determined) — so scripts can adapt their output to the available space
 - Displays command output inline in your status line
 - Configurable timeout (default: 1000ms)
 - Optional max-width truncation
@@ -217,7 +218,7 @@ Execute shell commands and display their output dynamically:
 
 > ⚠️ **Important:** Commands should complete quickly to avoid delays. Long-running commands will be killed after the configured timeout. If you're not seeing output from your custom command, try increasing the timeout value (press 't' in the editor).
 
-> 💡 **Tip:** Custom commands can be other Claude Code compatible status line formatters. They receive the same JSON via stdin that `ccstatusline` receives from Claude Code, allowing you to chain or combine multiple status line tools.
+> 💡 **Tip:** Custom commands can be other Claude Code compatible status line formatters. They receive the same JSON via stdin that `ccstatusline` receives from Claude Code (augmented with a `terminal_width` field), allowing you to chain or combine multiple status line tools.
 
 ### Link Widget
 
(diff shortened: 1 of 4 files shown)


sirmalloc:

PR #409, opened 2026-06-01
feat(widget): add cache-hit-rate widget

sebryu:
## Summary

Adds a `cache-hit-rate` widget that displays the prompt-cache hit rate for the current session, using the same formula the Anthropic Console reports.

Formula:

```
hit_rate = cache_read / (cache_read + cache_creation + fresh_input)
```

Renders as `Cache: 87%` (or `87%` in raw-value mode). Returns `null` until any cache or fresh-input tokens are seen, so it does not flash `0%` on empty transcripts.

## Why

Prompt-cache effectiveness is one of the most useful signals for tuning Claude Code workflows — high hit rate means cheaper, faster turns. It is visible on console.anthropic.com but there is no way to see it at a glance during a session. This widget surfaces the same number directly in the status line.

## Changes

- `src/widgets/CacheHitRate.ts` — new widget, follows the existing token-widget pattern (raw-value support, colors, preview)
- `src/widgets/index.ts`, `src/utils/widget-manifest.ts` — register `cache-hit-rate`
- `src/types/TokenMetrics.ts` — adds two **optional** fields: `cacheReadTokens` and `cacheCreationTokens`. The existing `cachedTokens` (read + creation) is preserved unchanged, so all existing widgets, configs, and downstream consumers are unaffected
- `src/utils/jsonl-metrics.ts` — sums `cache_read_input_tokens` and `cache_creation_input_tokens` separately while still rolling them into `cachedTokens`
- Tests:
  - `src/widgets/__tests__/CacheHitRate.test.ts` — widget behaviour (null guards, formula, raw mode, preview)
  - `src/utils/__tests__/jsonl-metrics.test.ts` — per-field aggregation in the existing fixtures

Net diff: +177 / -3 across 7 files (3 of those are 1-line registrations/exports).

## Compatibility

- `TokenMetrics.cachedTokens` semantics unchanged
- New fields on `TokenMetrics` are optional
- No changes to public CLI surface or settings format
- New widget is opt-in via the TUI

## Test plan

- [x] `bun run lint` passes (`tsc --noEmit` + ESLint `--max-warnings=0`)
- [x] `bun test` — all 1401 tests pass, including 7 new widget tests and 6 added assertions in existing jsonl-metrics fixtures
- [x] Manual: rendered against a real transcript piped through `bun run src/ccstatusline.ts` and verified the percentage matches what Anthropic Console reports for the same session

### DIFF
--- a/src/types/TokenMetrics.ts
+++ b/src/types/TokenMetrics.ts
@@ -17,6 +17,8 @@ export interface TokenMetrics {
     inputTokens: number;
     outputTokens: number;
     cachedTokens: number;
+    cacheReadTokens?: number;
+    cacheCreationTokens?: number;
     totalTokens: number;
     contextLength: number;
 }
--- a/src/utils/__tests__/jsonl-metrics.test.ts
+++ b/src/utils/__tests__/jsonl-metrics.test.ts
@@ -156,6 +156,8 @@ describe('jsonl transcript metrics', () => {
             inputTokens: 1799,
             outputTokens: 141,
             cachedTokens: 92,
+            cacheReadTokens: 56,
+            cacheCreationTokens: 36,
             totalTokens: 2032,
             contextLength: 250
         });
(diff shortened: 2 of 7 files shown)


sirmalloc:
Thanks for this, I'll publish it in the next release.

PR #411, opened 2026-06-02
fix: honour flex-separator in powerline render path

ssoriche:
## Summary

`renderPowerlineStatusLine` filters out `flex-separator` widgets at the top of the function and never restores their layout effect, so a configured `flex-separator` silently has no effect when powerline mode is enabled. The non-powerline render path correctly distributes terminal-width minus content-width across each `flex-separator` position; this PR teaches the powerline path to do the same.

### Repro

In an existing config, enable `powerline.enabled: true` and place a `flex-separator` widget between two regular widgets. Render against a known terminal width and observe the visible width of the result is content-width, not terminal-width — the flex space is missing.

### Fix

In `src/utils/renderer.ts`, inside `renderPowerlineStatusLine`:

1. **Track flex positions before filtering.** Build a `Set<number>` of `filteredWidgets` indices that are immediately followed by a `flex-separator` in the original widgets array. The widget itself stays filtered (powerline can't render it as a triangle), but the layout intent is preserved.
2. **Emit a sentinel + section cap during the render loop.** When a flex position is reached, emit a closing powerline cap — a triangle in the previous widget's bg colour with no bg of its own, so it terminates the left section visually — followed by a `FLEX_SEP` sentinel. Skip the regular between-widgets separator at that boundary.
3. **Post-process after the loop.** If any flex positions were marked, split the result on the sentinel, measure each part's visible width, distribute `terminalWidth − totalContentWidth` evenly across the flex positions as spaces, and reassemble. When the terminal width is unknown the sentinels are stripped so they can't leak into the output.

The non-powerline path is unchanged.

## Test plan

Four new tests in `src/utils/__tests__/renderer-flex-width.test.ts`:

- `distributes remaining width across a flex-separator in powerline mode` — single flex, asserts visible width equals the effective render width (`terminalWidth − 6` for `flexMode: 'full'`).
- `distributes space across multiple flex-separators in powerline mode` — three sections separated by two flex-separators, total visible width still matches effective width.
- `strips flex-separator markers when terminal width is unknown` — sentinel never leaks into output when no width is detected.
- `still works in non-powerline mode (no regression)` — sanity check on the existing path.

- [x] All 585 tests pass (`bun test`) — 581 existing + 4 new
- [x] TypeScript checks clean (`bun tsc --noEmit`)
- [x] ESLint clean (`bun run lint`)
- [x] Manually verified in a real Claude Code session (visible width fills to `terminalWidth − 6` as expected)


sirmalloc:
Thanks for this, it's live now.  I made several changes and bug fixes.  Main thing was allowing flex separators to be added while in powerline mode.  There were some start / end cap glitches, and I had it treat each segment of powerline widgets as a unique piece that continued the start / end cap assignment.  It feels pretty intuitive now and you can build some cool layouts.  I added a screenshot of one to the docs.

Issue #414, opened 2026-06-04
Feature request: widget to display extra usage used_credits


mhriemers:
## Summary

The usage API's `extra_usage.used_credits` is already fetched and cached by ccstatusline (`extraUsageUsed` in `CachedUsageDataSchema`), but no widget can display it. Please add an `extra-usage-used` widget — or have `extra-usage-remaining` fall back to showing the used amount when `extraUsageLimit` is absent.

## Motivation

For accounts with extra usage enabled but **no monthly limit configured** (`monthly_limit: null` — pay-as-you-go with no cap), neither existing extra-usage widget can render anything:

| Widget | Shows | Needs |
|---|---|---|
| `extra-usage-remaining` | limit − used | `extraUsageLimit` — absent |
| `extra-usage-utilization` | used / limit % | `extraUsageUtilization` — absent |

`used_credits` is the only meaningful extra-usage number for such accounts, and arguably interesting for capped accounts too (a used/remaining toggle, similar in spirit to #391).

## Current workaround

A `custom-command` widget reading ccstatusline's own cache:

```sh
jq -r '(.extraUsageUsed // 0) / 100' "$HOME/.cache/ccstatusline/usage.json" | LC_ALL=C xargs printf 'extra: €%.2f'
```

Related: #413 (these widgets currently show a permanent `[Timeout]` for limit-less accounts), #415 (currency formatting).

PR #416, opened 2026-06-04
fix: permanent [Timeout] in extra usage widgets when no monthly limit is set

### LINKED ISSUE #413: Extra usage widgets show permanent [Timeout] when account has no monthly limit set
## Summary

For accounts that have extra usage **enabled but no monthly limit configured** (`monthly_limit: null`), both extra-usage widgets permanently render `[Timeout]` even though the usage API responds quickly and successfully.

## Environment

- ccstatusline 2.2.19 (npx, auto-update), macOS 15
- Claude subscription with extra usage enabled, **no monthly limit set**

## Raw API response (`/api/oauth/usage`, trimmed)

```json
"extra_usage": {
  "is_enabled": true,
  "monthly_limit": null,
  "used_credits": 542.0,
  "utilization": null,
  "currency": "EUR",
  "disabled_reason": null
}
```

## What happens and why

`USAGE_WIDGET_REQUIREMENTS` declares `extra-usage-remaining` as requiring `extraUsageEnabled`, `extraUsageLimit`, and `extraUsageUsed` (and `extra-usage-utilization` requires `extraUsageUtilization`). With `monthly_limit: null` the API never supplies `extraUsageLimit`/`extraUsageUtilization`, so `hasRequiredUsageFields()` fails on every fetch — including perfectly successful ones. The fetch loop treats the response as incomplete, re-fetches each cycle, and the widget falls through to `getUsageErrorMessage()`, displaying **`[Timeout]`** indefinitely.

That error string is doubly misleading: the network round-trip succeeded, and no amount of waiting/cache-clearing can ever resolve it — the field will simply never exist for this account configuration.

Note `hasRequiredUsageField()` already special-cases `extraUsageEnabled === false` to excuse the detail fields; the `enabled === true && monthly_limit === null` configuration (pay-as-you-go with no cap) just isn't accounted for.

## Expected behavior

When the fetch succeeds but a widget's required field is absent from the response, render something honest (`n/a`, or nothing) instead of a network-error string, and treat the response as complete so it doesn't trigger continuous re-fetching.

Related: #414 (feature request: widget for `used_credits` — the only displayable extra-usage number for limit-less accounts), #415 (hardcoded `$` vs the API's `currency` field).

Happy to provide more details or test a fix.

mhriemers:
## Summary

Fixes #413 — accounts with extra usage **enabled but no monthly limit configured** (`monthly_limit: null`) saw the extra-usage widgets render a permanent `[Timeout]` even though every API fetch succeeded.

## Root cause

`hasRequiredUsageField()` treated the absence of `extraUsageLimit`/`extraUsageUtilization` as incomplete data. For limit-less accounts those fields never appear in the API response, so:

1. cached data (memory + file) was never accepted for widgets requiring them,
2. the fetch loop re-fetched every cycle, and
3. `getStaleUsageOrError()` surfaced the lock error as `[Timeout]` — indefinitely, since no amount of refetching can produce a field the account doesn't have.

## Fix

Extend the existing `extraUsageEnabled === false` special case to any response where the extra usage state is known: once the API has reported `is_enabled`, absent detail fields are conclusive.

With this change the widgets render nothing for limit-less accounts (instead of a false network error), matching how other widgets handle genuinely unavailable data. #414 proposes showing `used_credits` for these accounts.

## Testing

- New regression test `treats enabled extra usage without a monthly limit as complete for extra usage widget fields`, with a fixture mirroring a real `monthly_limit: null` API response — fails without the fix (cache rejected, second probe re-fetches), passes with it
- Full suite: 1394 tests pass, `bun run lint` clean


mhriemers:
Pushed a small test tweak: removed `currency: 'EUR'` from the new fixture so this PR and #418 stay green regardless of merge order (#418 starts parsing that field, which would have added `extraUsageCurrency` to the parsed result and broken this test's expectation).

sirmalloc:
Thanks for this, I'll publish it in the next release

PR #418, opened 2026-06-04
fix: format extra usage in the currency reported by the usage API

### LINKED ISSUE #415: Extra usage widgets hardcode $ instead of using the API's currency field
## Summary

The usage API reports the account's billing currency (`"currency": "EUR"` in my case), but `ExtraUsageRemaining` formats with a hardcoded `$`. Once a monthly limit is set, an EUR account sees its euros displayed with a dollar sign.

## Raw API response (`/api/oauth/usage`, trimmed)

```json
"extra_usage": {
  "is_enabled": true,
  "monthly_limit": null,
  "used_credits": 542.0,
  "utilization": null,
  "currency": "EUR",
  "disabled_reason": null
}
```

## Suggested fix

Parse the `currency` field (it isn't currently in `UsageApiResponseSchema`/`CachedUsageDataSchema`) and format with the matching symbol, falling back to `$`/the raw currency code when unknown.

Related to but distinct from #387, which fixed the cents/dollars math. Also related: #413, #414.

mhriemers:
## Summary

Fixes #415 — the usage API reports the account's billing currency (`extra_usage.currency`, e.g. `"EUR"`), but `ExtraUsageRemaining` always formatted with a hardcoded `$`, so non-USD accounts see their budget with the wrong symbol.

## Changes

- **Parse the field end-to-end**: `currency` added to `UsageApiResponseSchema`, `CachedUsageDataSchema`, both parse functions, `USAGE_DATA_FIELDS` / `pickDefinedUsageFields` in the prefetch merge, and the `UsageData` / `RenderUsageData` types (as `extraUsageCurrency`)
- **New shared formatter** `src/widgets/shared/currency.ts`: `formatUsageCurrency()` uses `Intl`-based currency formatting with the `en-US` locale, falling back to USD when the code is absent or invalid — so **output for USD accounts is byte-for-byte unchanged** (`$3,894.00`)
- `ExtraUsageRemaining` formats with the reported currency: an EUR account now sees `Overage Left: €3,894.00`

## Testing

- New `currency.test.ts` (default, known codes, invalid-code fallback)
- New EUR case in `ExtraUsageRemaining.test.ts`
- `currency: 'EUR'` added to the usage-fetch API fixture, asserting it parses through to `extraUsageCurrency`
- Full suite: 1397 tests pass, `bun run lint` clean

Related to but distinct from #387 (cents/dollars math). Related: #413/#416, #414/#417.

### DIFF
--- a/src/types/RenderContext.ts
+++ b/src/types/RenderContext.ts
@@ -20,6 +20,7 @@ export interface RenderUsageData {
     extraUsageLimit?: number;
     extraUsageUsed?: number;
     extraUsageUtilization?: number;
+    extraUsageCurrency?: string;
     error?: 'no-credentials' | 'timeout' | 'rate-limited' | 'api-error' | 'parse-error';
 }
 
--- a/src/utils/__tests__/usage-fetch.test.ts
+++ b/src/utils/__tests__/usage-fetch.test.ts
@@ -309,7 +309,8 @@ describe('fetchUsageData error handling', () => {
             is_enabled: true,
             monthly_limit: 400000,
             used_credits: 10600,
-            utilization: 2.6
+            utilization: 2.6,
+            currency: 'EUR'
         }
     });
     const rateLimitedResponseBody = JSON.stringify({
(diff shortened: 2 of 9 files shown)


sirmalloc:

Issue #420, opened 2026-06-06
[Windows] Process hangs indefinitely when spawned as child subprocess via Volta shim (stdin EOF not propagated through cmd.exe → bash → volta → node chain)


FelixIsaac:
## Summary

When `ccstatusline` is invoked as a child process from another Node.js script using `spawn('ccstatusline', [], { shell: true })` on Windows with a Volta-managed installation, the process hangs indefinitely and never exits. Each invocation leaks one `node.exe` process. In a long Claude Code session this accumulated to **961 leaked processes → 2.1 GB RAM → 90.5% kernel-mode CPU** from Windows scheduler overhead.

## Root cause

The Volta-installed `ccstatusline` binary is a bash shim:
```bash
volta run "$(basename $0)" "$@"
```

When spawned via `spawn('ccstatusline', [], { shell: true })`, the process chain on Windows is:

```
parent node → cmd.exe (shell:true) → bash (Volta shim) → volta → node (ccstatusline)
```

`ccstatusline` reads stdin with `for await (const chunk of process.stdin)` (line ~77342 in dist). The parent calls `cc.stdin.end()` to signal EOF — but on Windows, **stdin EOF does not propagate through the `cmd.exe → bash → volta` chain**. The inner `for await` loop never terminates. ccstatusline sits waiting forever. The parent also hangs waiting on `cc.on('close', ...)`.

Claude Code's statusline lifecycle **abandons** (does not kill) a hung process when a new refresh triggers, so processes accumulate indefinitely.

## This is a known Node.js/Windows issue

- nodejs/node#5620 — `process.stdin` never returns in Git Bash/Windows  
- nodejs/node#10836 — canonical Windows stdin hang  
- nodejs/node#32291 — piping `process.stdin` to `child.stdin` leaves open handle  
- nodejs/node#52364 — `spawn` + `stdin:'pipe'` causes pwsh to hang (still broken 2024)  
- volta-cli/volta#1199 — Volta shims + child_process confirmed broken on Windows  
- ryoppippi/ccusage#459 — **identical failure mode** in the same Claude Code ecosystem (statusline subprocess never exits, 300%+ CPU, 3+ GB RAM)

## The fix (2 lines in src/ccstatusline.ts)

After `renderMultipleLines` writes its output, instead of relying on event loop drain (which doesn't work through the Volta chain):

```typescript
// After rendering is complete:
process.stdout.write(formattedOutput, () => {
[...]

PR #421, opened 2026-06-06
fix: prevent process hang on Windows when stdin EOF does not propagate through Volta shim

FelixIsaac:
## Problem

On Windows with a Volta-managed ccstatusline, spawning ccstatusline as a child subprocess causes it to hang indefinitely. The process chain:

```
parent node → cmd.exe (shell:true) → bash (Volta shim) → volta → node (ccstatusline)
```

The `for await (const chunk of process.stdin)` loop never terminates because stdin EOF from the parent does not propagate through `cmd.exe → bash → volta` on Windows. Claude Code's statusline lifecycle **abandons** (does not kill) processes that take too long — so each refresh leaks one hung node.exe process.

**Real-world impact:** 961 leaked node.exe processes, 2.1 GB RAM, 90.5% kernel-mode CPU from Windows scheduler overhead managing ~1000 sleeping processes.

This is a known Windows/Node.js behavior:
- nodejs/node#32291 — piping stdin to child.stdin leaves open handle
- nodejs/node#52364 — spawn + stdin:pipe causes pwsh to hang (2024)
- volta-cli/volta#1199 — Volta shims + child_process broken on Windows cmd
- ryoppippi/ccusage#459 — identical failure mode in Claude Code ecosystem (same fix applied)

Note: the normal use case (`"command": "ccstatusline"` direct from Claude Code) is **not affected** — Claude Code pipes directly, single process layer, EOF works correctly.

## Fix

**Fix 1 — `readStdin()`:** Replace `for await` with event-based reading + 3s bail timeout. When EOF never arrives (Volta chain), the timeout destroys stdin and resolves the Promise. Data already accumulated in `chunks[]` is preserved and returned normally — no data loss.

**Fix 2 — `main()`:** Add explicit `process.exit(0)` after `renderMultipleLines`. Claude Code's "abandon not kill" lifecycle means any residual async handles from rendering keep the process alive indefinitely. An explicit exit prevents accumulation.

## Build note

Bun was not available in this environment so `dist/` was not rebuilt. The source change is in `src/ccstatusline.ts` only — maintainer build required before publishing.

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -65,11 +65,21 @@ async function readStdin(): Promise<string | null> {
                 chunks.push(decoder.decode(chunk));
             }
         } else {
-            // Node.js environment
+            // Node.js environment.
+            // Use event-based reading instead of for-await so we can add a bail
+            // timeout: on Windows, stdin EOF never propagates through the Volta shim
+            // chain (cmd.exe -> bash -> volta -> node), causing for-await to hang
+            // indefinitely. The timeout destroys stdin after 3s of silence; data
+            // already in chunks[] is preserved and returned normally.
+            // See: nodejs/node#32291, volta-cli/volta#1199, ryoppippi/ccusage#459
             process.stdin.setEncoding('utf8');
-            for await (const chunk of process.stdin) {
-                chunks.push(chunk as string);
-            }
+            await new Promise<void>((resolve) => {
+                const bail = setTimeout(() => { process.stdin.destroy(); resolve(); }, 3000);
+                if (typeof (bail as NodeJS.Timeout).unref === 'function') (bail as NodeJS.Timeout).unref();
+                process.stdin.on('data', (chunk: string) => chunks.push(chunk));
+                process.stdin.on('end', () => { clearTimeout(bail); resolve(); });
+                process.stdin.on('error', () => { clearTimeout(bail); resolve(); });
+            });
         }
         return chunks.join('');
     } catch {
@@ -292,6 +302,10 @@ async function main() {
                 }
 
                 await renderMultipleLines(result.data);
+                // Explicitly exit so the event loop does not linger. Without this,
+                // any residual async handles from rendering keep the process alive,
+                // which accumulates under Claude Code's "abandon not kill" lifecycle.
+                process.exit(0);
             } catch (error) {
                 console.error('Error parsing JSON:', error);
                 process.exit(1);


CorticalCode:
The root-cause diagnosis here is correct and bounding the read is the right direction. A few edge cases worth resolving before this lands, since this same `readStdin()` change is also the fix for the general (non-Volta) reports in #420 (see @mrns's no-Volta repro) and #485:

**Scope**
- Only the Node branch is guarded. The Bun branch (`src/ccstatusline.ts:66-71`) still uses `for await (const chunk of Bun.stdin.stream())` with no timeout, so `bun run` / from-source runs still hang indefinitely. The new `process.exit(0)` can't help there because execution is blocked inside `readStdin()` before render is reached. The published binary targets Node so end users are covered, but the two paths now diverge.

**Timeout semantics**
- The 3s timer is set once and cleared only on `end`/`error`; it is never reset on `data`, so it acts as a hard wall-clock cap rather than an idle timeout. Input that arrives in chunks crossing the 3s boundary gets `destroy()`'d mid-stream and then fails `JSON.parse` at `src/ccstatusline.ts:304`. Resetting the timer on each `data` (or resolving as soon as a complete JSON value/line is in hand, which is what the `head -1` workaround does) avoids both the truncation and the fixed 3s wait that every leaked process now incurs even when the JSON arrived immediately.

**Exit vs stdout flush**
- `process.exit(0)` immediately after `renderMultipleLines` can terminate before piped stdout has flushed. Under Claude Code stdout is a pipe, where Node writes may be asynchronous, and `process.exit()` is documented to abandon pending stdout/stderr I/O. With several `console.log()` writes (`src/ccstatusline.ts:213`, `:237`) this risks an intermittently truncated or blank status line. A flush-safe approach (destroy stdin to unblock the loop and let the event loop drain naturally, or await stdout before exiting) avoids the race. Worth noting the natural-EOF path does not need a forced exit at all; only the no-EOF/timeout path does.

**Tests**
- Coverage for stdin held open without EOF, slow/chunked input, the Bun branch, and a non-truncated-output assertion would keep this from regressing later.

FelixIsaac:
Thank you for the thorough review and clear pointers, @CorticalCode! All points have been addressed in commit e497d6e:

### 1. Scope (Bun Branch)
- Extracted the stdin reading logic into `src/utils/stdin.ts`, adding `readStdinBun()` alongside `readStdinNode()`.
- `readStdinBun()` consumes `Bun.stdin.stream()` via a `ReadableStreamDefaultReader`, racing each chunk read against an idle timeout and cancelling the reader/releasing the lock on completion. From-source / `bun run` executions now handle missing EOF identically to the bundled Node binary.

### 2. Timeout Semantics (Idle Timeout & Early JSON Resolution)
- Replaced the fixed wall-clock timeout with an idle timer that resets on each incoming chunk. Chunked or slow inputs crossing timeout intervals will no longer be destroyed mid-stream.
- Added `isCompleteJson()` check: as soon as a complete, valid JSON object is received, `readStdin()` resolves immediately. Under the Windows/Volta shim where EOF never propagates, status line renders now complete instantly (a few milliseconds) rather than waiting out a timeout.

### 3. Exit vs. stdout Flush
- Removed the immediate `process.exit(0)` after `renderMultipleLines()`, which risked truncating piped stdout under backpressure.
- Added a `flushStdout()` helper that checks `stdout.writableNeedDrain` / `stdout.writableLength` and awaits the `drain` event.
- Stdin streams are explicitly destroyed/cancelled in `readStdin()` once reading completes, unblocking the event loop so it drains naturally once stdout flushes.
- Added an unreferenced fallback timer (1s) after `flushStdout()` to prevent any rogue third-party async handles from keeping the process alive without risking stdout truncation.

### 4. Test Coverage
- Added a full test suite in `src/utils/__tests__/stdin.test.ts` (20 tests passing) covering:
  - `isCompleteJson()` parsing edge cases (single-line, multi-line, nested, fragments, braces inside strings, non-JSON).
  - Both Node (`PassThrough`) and Bun (`ReadableStream`) streams with normal EOF, held-open streams without EOF, and chunked streams across idle timer intervals.
  - Stream destruction / reader cancellation on completion.
  - End-to-end subprocess tests for both Node (`dist/ccstatusline.js`) and Bun (`src/ccstatusline.ts`) with held-open stdin pipes, asserting clean exit code 0 in < 1s and complete, non-truncated statusline output.
- Verified clean build (`bun run build`) and type check / linter passes with zero errors/warnings (`bun run lint`).

PR #424, opened 2026-06-09
Fix git-review detection with SSH host aliases

### LINKED ISSUE #419: git-review shows (no PR) when origin uses an SSH host alias (e.g. git@myalias:owner/repo)
## Summary

The `git-review` widget shows `(no PR)` when the `origin` remote uses an **SSH host alias** (from `~/.ssh/config`) instead of a literal `github.com` / `gitlab.com` host. A bare `gh pr view` resolves such remotes correctly, so the PR is detectable - ccstatusline's own provider-selection logic is what fails.

## Version

ccstatusline 2.2.19

## Repro

`~/.ssh/config`:

```
Host mygit
	HostName github.com
	User git
	IdentityFile ~/.ssh/some_key
	IdentitiesOnly yes
```

Repo remote:

```
$ git remote get-url origin
git@mygit:owner/repo.git
```

There is an open PR for the current branch. Confirm it is detectable normally:

```
$ gh pr view --json number,state,title
{"number":1485,"state":"OPEN","title":"..."}     # works
```

But the status line shows `(no PR)`.

## Root cause

The remote URL host is parsed **literally** as `mygit` (an SSH alias), and `getProviderCandidates()` never selects a provider:

```js
const host = getOriginHost(...);          // "mygit"
if (host.includes("github")) return ["gh"];   // miss
if (host.includes("gitlab")) return ["glab"]; // miss
// fallback: gate on `<cli> auth status --hostname mygit`
```

```
$ gh auth status --hostname mygit
You are not logged into any accounts on mygit
```

Both candidates fail that gate, so `getProviderCandidates()` returns `[]`, the loop in `fetchGitReviewData` runs zero times, and the result is cached as `null` -> `(no PR)`. `gh` is never invoked at all.

Note: `fetchFromProvider` *already* tries an unpinned bare `gh pr view` (no `--repo`) before falling back to the alias-derived `--repo https://mygit/...`, and that bare call works fine with alias remotes. So the only thing standing between this setup and a correct result is provider selection - if `gh` were in the candidate list, the existing unpinned path would succeed. (The alias-based `--repo` would also fail with `error connecting to mygit`, but that path is never reached here.)

## Suggested fix

Resolve SSH host aliases to their real `HostName` before host matching (and before building `--repo`). `ssh -G <host>` does this and is a safe no-op for literal hosts:

```
$ ssh -G mygit      | grep '^hostname '   ->  hostname github.com
$ ssh -G github.com | grep '^hostname '   ->  hostname github.com
```

A small helper applied in `getOriginHost` (so provider selection sees `github.com`) - and optionally in `getOriginRepoRef` so the pinned `--repo` is also correct - fixes it, degrading to the original host when `ssh` is unavailable. Alternatively, when the origin host matches no known provider, fall back to trying available CLIs via the unpinned path instead of gating on `auth status --hostname <alias>`.

cyphercodes:
## Summary
- resolve SSH host aliases with `ssh -G` before git-review provider selection
- use the resolved host when constructing fallback `--repo` refs
- add coverage for an origin like `git@mygit:owner/repo.git` resolving to GitHub

Fixes #419.

## Tests
- `bun test src/utils/__tests__/git-review-cache.test.ts`
- `bun run lint`
- `git diff --check`

### DIFF
--- a/src/utils/__tests__/git-review-cache.test.ts
+++ b/src/utils/__tests__/git-review-cache.test.ts
@@ -377,6 +377,40 @@ describe('git-review-cache', () => {
         expect(ghPrCalls[1]?.args).toContain('https://github.com/owner/repo');
     });
 
+    it('preserves canonical GitHub SSH hosts when SSH config points at a transport endpoint', () => {
+        const harness = createHarness();
+        harness.setOriginRemoteUrl('git@github.com:owner/repo.git');
+        harness.setSshHostAlias('github.com', 'ssh.github.com');
+        harness.ghResponses.push('');
+        harness.ghResponses.push(JSON.stringify({
+            number: 1486,
+            reviewDecision: '',
+            state: 'OPEN',
+            title: 'Canonical GitHub PR',
+            url: 'https://github.com/owner/repo/pull/1486'
+        }));
+
+        expect(fetchGitReviewData('/tmp/repo', harness.deps)).toEqual({
+            number: 1486,
+            provider: 'gh',
+            reviewDecision: '',
+            state: 'OPEN',
+            title: 'Canonical GitHub PR',
+            url: 'https://github.com/owner/repo/pull/1486'
+        });
+
+        const sshCalls = harness.execCalls.filter(call => call.cmd === 'ssh');
+        expect(sshCalls).toHaveLength(0);
+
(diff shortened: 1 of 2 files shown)


sirmalloc:

PR #425, opened 2026-06-10
fix(compaction): count compact_boundary markers instead of inferring from context-% drops

CorticalCode:
## Why

The Compaction Counter (#282) infers compactions from drops in `used_percentage` between renders. The heuristic is fragile: it samples a noisy, debounced signal and guesses about discrete events. Known failure: transient `used_percentage: 0` frames create false counts — #370 (thanks @KorenKrita) patches that symptom with a sub-1% guard, but the problem class is inherent to inference: missed events between renders, window-size changes, and persisted prev-state that can go stale.

## What

Count the explicit `{"type":"system","subtype":"compact_boundary"}` records Claude Code writes to the session transcript at every compaction. Exact, stateless, retroactive. Records with `isSidechain: true` (a subagent compacting its own context) are excluded, matching the main-chain convention already used in `jsonl-metrics.ts`. Display is unchanged (`↻ N`). The heuristic, its zod schema, threshold logic, and the per-session cache under `~/.cache/ccstatusline/compaction/` are deleted — net **−319 lines**. Marker format verified stable across transcripts from v2.1.92 → v2.1.170, for both `auto` and `manual` triggers (v2.1.170 adds extra `compactMetadata` fields, but `type`/`subtype`/`isSidechain` are unchanged).

## Behavior notes

- Stateless and file-scoped: concurrent Claude Code instances count independently, and the count accumulates across `--resume`/`--continue` (markers live in the session's transcript file).
- Gated on `transcript_path`, like every other transcript metric.
- Adds one transcript read per render when the widget is enabled — the same per-metric pattern as token/duration/speed metrics. (A shared single-read across metrics would be a nice follow-up.)
- Old per-session cache files under `~/.cache/ccstatusline/compaction/` become inert; they're left in place, consistent with how the existing per-session skills cache is handled.
- The 0%-flash false-count class can't occur by construction.

## Testing

- 10 unit tests (TDD): exact counting, 0%-flash immunity, sidechain exclusion, malformed lines, missing/unreadable transcripts.
- Full suite: 1369 pass / lint clean / bundled build verified.
- End-to-end: piped renders against fixtures and real-world transcripts, cross-checked against `grep -c` marker counts — old code shows a false `↻ 1` on a flash frame; this branch shows `↻ 0` (no markers) and exactly `↻ 2` (two markers + flashes + a sidechain marker).


sirmalloc:
Thanks, it'll go out in the next release

PR #432, opened 2026-06-13
fix: keep usage-fetch tests from touching the real user cache

### LINKED ISSUE #437: bun test corrupts the developer's real ccstatusline cache on Windows
Running `bun test` (the usage-fetch suite) on Windows writes fixture data and stale locks into the real `~/.cache/ccstatusline`. The probe subprocesses override `HOME` but inherit the real `USERPROFILE`, which `os.homedir()` prefers on Windows, so every probe escapes its sandbox home.

Symptoms: a live statusline shows `[Timeout]` until the leaked lock expires and fixture percentages until the cache ages out; all 11 tests in `src/utils/__tests__/usage-fetch.test.ts` also fail on Windows.

Repro: run `bun test src/utils/__tests__/usage-fetch.test.ts` on Windows while a statusline is active, then inspect `%USERPROFILE%\.cache\ccstatusline\usage.json`.

zachthedev:
Closes #437.

Running `bun test` corrupts the developer's live statusline on Windows: the usage-fetch probe subprocesses override `HOME` but inherit the real `USERPROFILE`, which `os.homedir()` prefers there. Every run wrote fixture data and stale locks into the real `~/.cache/ccstatusline`, so the running statusline showed `[Timeout]` until the lock expired and fixture percentages until the cache aged out. It also made all 11 tests in the file fail on Windows.

Fixes, all inside the test file:

- pin `USERPROFILE` alongside `HOME` in the probe env (the actual escape)
- pin `CLAUDE_CONFIG_DIR` / `HTTPS_PROXY` / `https_proxy` so inherited developer settings never reach probes
- every probe now reports `os.homedir()` and the harness asserts it matches the sandbox home, so a future isolation escape fails loudly instead of silently corrupting user state
- the lowercase `https_proxy` assertion is platform-aware: Windows env vars are case-insensitive, so it is indistinguishable from `HTTPS_PROXY` there

Windows: this file goes from 0/11 to 11/11 passing. POSIX behavior unchanged.

### DIFF
--- a/src/utils/__tests__/usage-fetch.test.ts
+++ b/src/utils/__tests__/usage-fetch.test.ts
@@ -21,6 +21,7 @@ interface UsageProbeResult {
     requestCount: number;
     proxyAgentConfigured: boolean;
     requestHost: string | null;
+    homedir: string;
     lockContents: string | null;
 }
 
@@ -150,6 +151,7 @@ process.stdout.write(JSON.stringify({
     requestCount,
     proxyAgentConfigured,
     requestHost,
+    homedir: os.homedir(),
     lockContents: fs.existsSync(lockFile) ? fs.readFileSync(lockFile, 'utf8') : null
 }));
 `;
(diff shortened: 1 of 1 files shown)


sirmalloc:
Thanks, I'll publish in the next release.

PR #433, opened 2026-06-13
feat: add per-widget dim styling, whole widget or parens-only

### LINKED ISSUE #438: Dim styling toggle (whole widget or just the parenthesized part)
Bold exists per widget, dim does not. The only workaround is baking darker hex colors into each widget, which has to be reverted by hand and cannot touch text rendered inside a widget, like the Context Bar's `(42%)`.

Ask: a (d)im toggle next to (b)old in the colors menu, with a mode that dims only `(...)` spans so the bracketed part of a widget can be de-emphasized while the value keeps full intensity.

zachthedev:
Closes #438.

Adds a per-widget dim toggle alongside bold. Press **(d)** in the colors menu to cycle: off → whole widget → `(...)` spans only → off. Parens mode dims just the bracketed part of a widget, e.g. the `(42%)` in Context Bar output, while the rest keeps full intensity.

| Area                                     | Change                                                                                                                                       |
| ---------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
| `types/Widget.ts`                        | `dim: z.union([z.boolean(), z.literal('parens')])` on widget items                                                                           |
| `utils/colors.ts`                        | `applyColors` gains a dim arg; `applyParensDim` wraps `(...)` spans and re-asserts bold after each span (`\x1b[22m` clears both intensities) |
| `utils/renderer.ts`                      | wired into normal and powerline paths; separators inherit dim like bold; intensity resets shared with bold                                   |
| `color-menu/mutations.ts`                | `cycleWidgetDim`; reset / clear-all strip `dim`                                                                                              |
| `ColorMenu.tsx`                          | (d) keybind, `[DIM]` / `[DIM ()]` indicator, dim shown in the widget list preview                                                            |
| `renderer-dim.test.ts` + mutations tests | applyColors dim modes, normal + powerline rendering, visible-width invariant, cycle behavior                                                 |

Notes:

- No settings migration needed: the field is optional and absent means off.
- Visible text and width are unchanged, so truncation and flex math are unaffected (covered by a test).
- Zero file overlap with my other open PRs; based on current main.

### DIFF
--- a/src/tui/components/ColorMenu.tsx
+++ b/src/tui/components/ColorMenu.tsx
@@ -23,6 +23,7 @@ import { ConfirmDialog } from './ConfirmDialog';
 import {
     clearAllWidgetStyling,
     cycleWidgetColor,
+    cycleWidgetDim,
     resetWidgetStyling,
     setWidgetColor,
     toggleWidgetBold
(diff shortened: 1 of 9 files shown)


sirmalloc:
Made a few tweaks, mostly due to interaction of dim and bold widgets next to each other, preview not resetting bold properly after a dim, issues with separators in powerline mode after a dim, and the UI wrapping the dim indicator to the next line in some instances when bold was also applied.  I'll publish it in the next release.

PR #434, opened 2026-06-13
fix: stop refetching usage every render on accounts without rate-limit windows

### LINKED ISSUE #343: Session usage not updating on Enterprise plan
## Description

After migrating from **Claude Max plan** to **Enterprise plan**, the status line appears but session usage data remains at zero / static — it never updates during the session.

Previously on Max plan, usage metrics updated in real-time as expected.

## Steps to Reproduce

1. Migrate Claude Code account from Max plan to Enterprise plan
2. Install/run ccstatusline via `npx -y ccstatusline@latest`
3. Start a Claude Code session and interact normally
4. Observe the status line — usage values remain at 0 or do not change

## Expected Behavior

The status line should display and update session usage metrics (tokens consumed, cost, usage vs plan limits) in real-time, reflecting the Enterprise plan's quota/limits — similar to how it worked on the Max plan.

## Actual Behavior

The status line renders but usage data stays zeroed out / static throughout the entire session. No errors are shown.

## Environment

| Component | Version |
|-----------|---------|
| ccstatusline | latest (via `npx -y ccstatusline@latest`) |
| Claude Code | 2.1.92 |
| Node.js | v24.14.0 |
| OS | Windows 10 Home (10.0.19045) |
| Plan | Enterprise (migrated from Max) |

## Additional Context

_Screenshots will be attached in a follow-up comment._

zachthedev:
Closes #343

Enterprise accounts have no 5-hour/7-day rate-limit windows, so the usage API returns `five_hour: null` and `seven_day*: null` (#343). Those buckets parse to utilization `0` with no `resets_at`, which leaves the cache permanently "incomplete" for any widget that requires a reset timestamp: `fetchUsageData` then bypasses the cache and hits the live API on **every render**. At a 15s refresh interval Anthropic starts answering 429 with `Retry-After: 300`, and the lock makes the usage/money widgets flicker `[Rate limited]` / `[Timeout]` indefinitely.

Observed on a real Enterprise account: the API answers 200 with all windows null, the cache holds `{"sessionUsage":0,...}` with no reset fields, and the lock sits at `{"blockedUntil":...,"error":"rate-limited"}` in a permanent refetch/backoff cycle.

| Area                  | Change                                                                                                                                                                                           |
| --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `usage-fetch.ts`      | a missing window reset field counts as satisfied once its paired utilization is cached: a null bucket parses to utilization 0 with no `resets_at`, so refetching can never produce the timestamp |
| `usage-fetch.test.ts` | probe test with the Enterprise response shape: first run fetches once, a second probe 10s later is served entirely from cache (zero requests)                                                    |

Notes:

- Same conclusiveness idea as #416, which covers the sibling `monthly_limit: null` case; the two PRs touch the same function and either rebases trivially on the other.
- No rendering changes; widgets display exactly what they did before, they just stop hammering the API.

### DIFF
--- a/src/utils/__tests__/usage-fetch.test.ts
+++ b/src/utils/__tests__/usage-fetch.test.ts
@@ -357,6 +357,23 @@ describe('fetchUsageData error handling', () => {
             disabled_reason: null
         }
     });
+    // Mirrors a real Enterprise-account response: every rate-limit window is
+    // null because Enterprise plans have no 5-hour/7-day windows (#343).
+    const enterpriseNullWindowsResponseBody = JSON.stringify({
+        five_hour: null,
+        seven_day: null,
+        seven_day_oauth_apps: null,
+        seven_day_sonnet: null,
+        seven_day_opus: null,
+        extra_usage: {
+            is_enabled: true,
+            monthly_limit: 50000,
+            used_credits: 0,
+            utilization: null,
+            currency: 'USD',
+            disabled_reason: null
+        }
+    });
     const rateLimitedResponseBody = JSON.stringify({
         error: {
             message: 'Rate limited. Please try again later.',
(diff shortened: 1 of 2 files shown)


sirmalloc:
@zachthedev If a user with an enterprise account logs out and logs back in with a personal account, will the usage ever refresh again if the  enterprise sentinel is written?

zachthedev:
> @zachthedev If a user with an enterprise account logs out and logs back in with a personal account, will the usage ever refresh again if the enterprise sentinel is written?

Good point, I can try with my setup.

zachthedev:
@sirmalloc 
> If a user with an enterprise account logs out and logs back in with a personal account, will the usage ever refresh again if the enterprise sentinel is written?

No. The sentinel only gates completeness (`hasRequiredUsageField`); the cache is served only when also fresh: `cacheAge < CACHE_MAX_AGE (180s) && hasRequiredUsageFields(...)`, ANDed. This PR touches only the second term, so a stale cache still refetches. After enterprise->personal it refetches at the 180s mark with the new token and the real windows replace it; the sentinel is recomputed per call (not persisted).

During that window it serves the prior snapshot, which is the existing <=180s lag of a non-account-keyed cache, not something the sentinel introduces.

#460 drops that 180s: it fingerprints the token into the cache and invalidates on mismatch. Orthogonal to this fix.

PR #435, opened 2026-06-13
fix: prefer cumulative transcript metrics for input/output token widgets

### LINKED ISSUE #439: Tokens In/Out no longer sum with Cached/Total on Claude Code >= 2.1.132
Claude Code 2.1.132 changed the statusline `context_window` token fields from cumulative session totals to current-context usage (changelog: "Fixed statusline `context_window` token counts reflecting cumulative session totals instead of current context usage").

Tokens In/Out prefer those fields (#168, added when they were still cumulative per #126), while Tokens Cached/Total always read the transcript. On Claude Code >= 2.1.132 the four-widget equation `in + out + cache = total` mixes current-context and cumulative values and stops summing, and In/Out can flash absurd mid-turn values (observed live: `Out: 3` next to a cumulative 445M cache) while contradicting their own "for the current session" descriptions.

zachthedev:
Closes #439.

Claude Code 2.1.132 changed the statusline `context_window` token fields from cumulative session totals to current-context usage (changelog: "Fixed statusline `context_window` token counts reflecting cumulative session totals instead of current context usage"). Tokens In/Out have preferred those fields since #168, written when they were still cumulative, so on CC ≥ 2.1.132 they contradict their own "for the current session" descriptions and the four-widget equation stops summing: `in + out + cache = total` mixes current-context and cumulative values (Cached/Total never read `context_window`).

| Area                                 | Change                                                                                                            |
| ------------------------------------ | ----------------------------------------------------------------------------------------------------------------- |
| `TokensInput.ts` / `TokensOutput.ts` | prefer `context.tokenMetrics` (transcript, cumulative on every CC version); `context_window` becomes the fallback |
| `TokensWidgets.test.ts`              | preference flipped, plus a fallback-direction test (`context_window` only, no transcript)                         |

Considered and rejected: keeping the new per-context semantics and adding separate context-window token widgets; that leaves In/Out contradicting their documented descriptions and the equation broken for existing configs. Happy to do that as a follow-up if you'd rather have both.


sirmalloc:
Thanks, I'll publish this in the next release.

PR #441, opened 2026-06-13
feat: total session tokens including sub-agents

elhoim:
## Summary

Adds **subagent-inclusive session token accounting**. The existing token widgets
count only the main agent — tokens spent by Task-tool sub-agents are invisible.
This wires sub-agent usage into token counting, reusing the same sub-agent
transcript discovery that already backs the speed-metrics widgets.

Two surfaces (opt-in, zero change for existing setups):

- **`(s)ubagents` toggle** on the existing **Tokens Input / Output / Cached / Total**
  widgets. When on, that widget's number includes sub-agents, shows a `Σ` marker,
  and a `[+sub]` modifier in the editor.
- **New `Session Total Tokens` widget** (`tokens-session-total`) — session grand
  total including sub-agents, with an optional `(b)reakdown` of input/output/cache:
  `Σ Total: 152k (in 90k/out 40k/cache 22k)`.

## How it works

- `getTokenMetrics(path, { includeSubagents })` sums referenced
  `subagents/agent-*.jsonl` files via the existing
  `getReferencedSubagentIds` → `getSubagentTranscriptPaths` discovery.
- **Default behavior is byte-identical** — the option defaults off; existing
  widgets and the stdin `context_window` path are unchanged.
- **Double-count guard:** when separate sub-agent files exist, inline
  `isSidechain` rows are dropped from the main pass (they're represented by the
  files); when no files exist (older transcript format), inline sidechain rows are
  still counted. `contextLength` remains a main-chain-only concept.
- The subagent-inclusive figure is computed **once per render, on demand** — only
  when a session-total widget or a subagents-enabled token widget is configured
  (`RenderContext.sessionTokenMetrics`). No extra file I/O otherwise.

## Testing

- New unit tests: data layer (incl. double-count guard, multi/unreferenced
  sub-agents, back-compat lock), the metadata helper, the four widget toggles, and
  the new widget. **49 tests pass** including the pre-existing `TokensWidgets`
  suite (no regression).
- `tsc --noEmit` and `eslint` clean across the project; `bun run build` succeeds;
  example-payload smoke render verified.


astral303:
Merge ed087b6 is bad, compilation errors starting in src/utils/jsonl-metrics.ts:256

elhoim:
Fixed it in [0f89739](https://github.com/sirmalloc/ccstatusline/pull/441/commits/0f89739d41addcdad792c9b7999e58a085ef3de9).

elhoim:
Rebuilt against current `main` — conflicts resolved and CI green.

The conflicts weren't textual: `main` replaced the array-based `getTokenMetrics()` this branch was written on with a single streaming pass (`scanTranscript`), and the token widgets picked up number formats and hideable zero states. So the feature is **re-ported**, not merged around:

- **`jsonl-metrics.ts`** — new `includeSubagentTokens` option puts `sessionTokenMetrics` on the analysis. Each sub-agent transcript is read **once** and feeds the token and speed collectors together, so the "computed once per render" promise still holds. Inline `isSidechain` usage now accumulates into its own tally and is subtracted from the session totals when separate sub-agent files exist — the old `skipMainSidechain` guard, in collector form. `contextLength` stays main-chain only.
- **`ccstatusline.ts`** — the subagent-inclusive pass is requested only when a widget actually asks for it, so nothing extra is read for existing configs.
- **Token widgets** — the `(s)ubagents` toggle, `Σ` marker and `[+sub]` modifier now sit on top of `resolveNumberFormat` and the zero-hideable state. With subagents on, the main-only `context_window` fallback is skipped.
- **Housekeeping** — `token-subagents.ts` and `SessionTotalTokens.ts` use the shared metadata helpers `main` introduced, and `Session Total Tokens` gained number-format support plus the zero-hideable state for parity with the rest of the Tokens category.

**Default behavior is unchanged:** without the opt-in, no sub-agent file is opened and every existing token number is identical.

Verification, beyond CI:

- The three subagent token tests were re-expressed against the new API (main metrics untouched, inline-sidechain guard, referenced-files-only) and pass alongside the existing widget tests.
- On a real session with 23 sub-agent transcripts, main metrics are unchanged (`in=334 out=318378 cached=27552149`) and the session totals (`in=810 out=519983 cached=47512932`) match an independent tally of that transcript plus its referenced agent files, to the token.
- Locally: `bun run lint` clean, `bun test` 2376 pass / 2 fail — both pre-existing flaky TUI/subprocess cases in my environment (`fetchUsageData` subprocess timeout, `UsageTimezoneEditor`), neither in metrics; CI is green on all three jobs.

One note for scheduling: #591 fixes a double-count in the same collector (one API response is logged once per content block, so every transcript-derived token number currently reads ~2x). The two are independent, but whichever lands second needs a trivial re-merge of `jsonl-metrics.ts`, and these session totals inherit that fix once it's in. Happy to rebase this on top of it in either order.

PR #445, opened 2026-06-15
feat(compaction): opt-in trigger split and tokens reclaimed

CorticalCode:
## Summary
Two opt-in, composable, per-item add-ons for the **Compaction Counter**, parsed from data already present in the `compact_boundary` transcript markers — no new data source, hook, cache, or dependency:
- `s` — **trigger split**: `↻ 3 (2 auto, 1 manual)`
- `t` — **tokens reclaimed**: `↻ 3 ↓887.0k`

The default `↻ N` and the existing `f` / `n` / `h` controls are unchanged; the two toggles are independent per-item and stack. Builds on #425 (marker-based counting), continuing the widget from #282.

## How it works
- `compaction.ts` returns `{ count, byTrigger, tokensReclaimed }` from the same single transcript scan it already performs. A missing or unrecognized trigger is bucketed as `unknown` (never guessed); `tokensReclaimed` sums each compaction's `preTokens − postTokens` (floored at 0, only when both fields are present), so older transcripts predating `postTokens` simply show nothing.
- The display add-ons reuse the shared `isMetadataFlagEnabled` / `toggleMetadataFlag` helpers, and `formatTokens` for the reclaimed figure.
- `formatTokens` was extracted into a small leaf module to avoid a `renderer → widget-registry → widget` import cycle; `renderer` re-exports it, so existing importers are unaffected.

## Test plan
- `bun test` green — unit + widget coverage (trigger buckets incl. `unknown`, reclaimed math/floor, sidechain exclusion, both toggles, stacking, preview, and the reclaimed figure near 1M).
- `bun run lint` clean; `bun run build` bundles.
- Verified end-to-end on a real transcript: toggles render and compose, subagent (sidechain) compactions excluded from every stat, default output unchanged.

## Note
`formatTokens` is also fixed by the separate small PR #444 (a rounding bug in `renderer.ts`). This PR's extracted `format-tokens.ts` carries the same one-line fix, so the reclaimed figure renders correctly near 1M and merging the two PRs in either order can't regress it — whichever lands second finds the fix already present.


sirmalloc:
Looks good, it'll go out in the next release.

PR #447, opened 2026-06-15
fix(config): non-destructive recovery + loud warning for invalid settings.json

### LINKED ISSUE #393: Configuration settings reset on invalid JSON
It took me some time to figure out but every time the  `settings.json` has invalid JSON, the settings reset to a default value and all your changes are lost. This is a bit annoying.

Offtopic: thank you very much for this amazing project 🙏

CorticalCode:
## Summary

Fixes #393. Today a single malformed `settings.json` silently wipes the user's entire
ccstatusline configuration. This PR makes configuration loading **non-destructive** — an
unreadable config is never overwritten — surfaces the failure **loudly** in the statusline
instead of swallowing it, and makes every config write **atomic**.

## The problem

`loadSettings()` runs on the statusline render path, i.e. on **every prompt**. Before this
PR, when `settings.json` failed to parse or validate, recovery backed the file up to
`settings.bak` and then **overwrote `settings.json` with defaults**. Because that happens
automatically on the render path, one bad edit — a stray comma, a trailing brace — silently
replaced the user's whole configuration with defaults. The only trace was a `console.error`
to stderr, which the piped statusline never shows. As #393 puts it: *"it took me some time
to figure out … all your changes are lost."*

## The fix

Three principles, applied to every path through config load/save:

1. **Never overwrite a config we can't read.** On any unreadable or invalid input, fall back
   to built-in defaults **in memory only** and leave the file exactly as it is on disk — so
   the user can correct the typo and their configuration returns on the next render, with no
   backup to hunt down and no restore step. (This extends a pattern already in the codebase:
   the legacy `git-pr` → `git-review` widget rewrite is likewise applied in memory on load
   and only persisted on the next save.)
2. **Fail loudly, not silently.** When recovery happens, the statusline prepends a red
   `⚠ invalid config` badge, so the problem is obvious instead of masquerading as a working
   (default) bar. The detailed reason still goes to stderr.
3. **Write atomically.** Config writes now go through a temp file + `rename` — the same idiom
   already used by `writePersistentCache()` in `git.ts` — so a reader on the render path can
   never observe a half-written file (which itself used to trip the reset).

## Behavior, before and after

| Scenario | Before this PR | After this PR |
|----------|----------------|---------------|
| File missing (first run) | write defaults | write defaults *(now atomic)* |
| Valid current config | load and use | load and use |
| Valid but older version | migrate → **write** → validate | migrate → validate → **write only if valid** |
| Invalid JSON | `.bak` + **overwrite with defaults** | **defaults in memory; file untouched** + `⚠` badge |
| Schema-invalid | `.bak` + **overwrite with defaults** | **defaults in memory; file untouched** + `⚠` badge |
| Unreadable file | `.bak` + **overwrite with defaults** | **defaults in memory; file untouched** + `⚠` badge |
| Migration yields an invalid result | **overwrites with the bad migration** | **defaults in memory; file untouched** + `⚠` badge |
| Recording install metadata over an unreadable file | overwrite with defaults + metadata | **skip the write; file untouched** |
| Any write (defaults / migration / save) | direct `writeFile` (can tear) | temp file + atomic `rename` |

On an invalid config the statusline now reads:

```
⚠ invalid config | Model: Opus 4.6 | Ctx: 15.5k | ⎇ main
```

## A note on the removed `.bak`

The previous recovery backed the bad file up to `settings.bak` before overwriting it. Since
recovery no longer overwrites anything, that backup would just be a copy of the original —
which is now left in place — so it has been removed. Happy to restore a `.bak` snapshot if
you'd prefer to keep one.

## Testing

- New/updated unit tests in `config.test.ts` and a new `renderer-config-warning.test.ts`
  cover: every recovery path preserving the file on disk (and writing no `.bak`); the
  migration-validates-before-persist guard; the install-metadata guard; atomic writes leaving
  no temp residue (and cleaning up on a failed rename); and the warning badge (red, and plain
  at color level 0).
- **Cross-platform matrix:** the suite was run on a **Linux / macOS / Windows** matrix. It is
  fully green on Linux and macOS, and every test exercising this change — including the atomic
  temp-file + `rename` path — passes on Windows as well. **This change introduces no new test
  failures on any platform.**

Closes #393

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -14,6 +14,7 @@ import { getVisibleText } from './utils/ansi';
 import { updateColorMap } from './utils/colors';
 import { getCompactionCount } from './utils/compaction';
 import {
+    getConfigLoadError,
     initConfigPath,
     loadSettings,
     saveSettings
@@ -26,6 +27,7 @@ import {
 } from './utils/jsonl';
 import { advanceGlobalPowerlineThemeIndex } from './utils/powerline-theme-index';
 import {
+    buildConfigWarningBadge,
     calculateMaxWidthsFromPreRendered,
     preRenderAllWidgets,
     renderStatusLine
@@ -88,6 +90,7 @@ async function ensureWindowsUtf8CodePage() {
 
 async function renderMultipleLines(data: StatusJSON) {
     const settings = await loadSettings();
+    const configError = getConfigLoadError();
 
     // Set global chalk level based on settings
     chalk.level = settings.colorLevel;
@@ -171,6 +174,7 @@ async function renderMultipleLines(data: StatusJSON) {
     // Render each line using pre-rendered content
     let globalSeparatorIndex = 0;
     let globalPowerlineThemeIndex = 0;
+    let configBadgePrepended = false;
     for (let i = 0; i < lines.length; i++) {
         const lineItems = lines[i];
         if (lineItems && lineItems.length > 0) {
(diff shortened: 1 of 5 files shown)


sirmalloc:
I merged this, and found one bug after merging relating to following symlinks that I'm going to resolve directly against main.  Thanks for the contribution.

CorticalCode:
Thanks for catching this and turning it around so fast. Appreciate you addressing it, and the merge 🙏

PR #451, opened 2026-06-15
fix(context-bar): render counts >= 1M as "1.0M" instead of "1000k"

### LINKED ISSUE #449: context-bar: counts >= 1M render as 1000k instead of 1.0M
`ContextBar` (`src/widgets/ContextBar.ts`) formats used/total as `Math.round(n / 1000) + "k"`, so a 1M context window renders as `1000k` (e.g. `504k/1000k`). `formatTokens` already rolls values >= 999950 up to `M`, and `ContextLength` uses it; the bar is the only token readout that does not.

Request: roll the bar used/total up to `M` at >= 1M (e.g. `711k/1.0M`), keeping the compact `k` form below that.

zachthedev:
Closes #449.

`ContextBar` formatted used/total with `Math.round(n / 1000) + 'k'`, so a 1M context window rendered as `1000k` (e.g. `504k/1000k`) — the only token readout ignoring the `M` rollup `formatTokens` already does.

`formatTokens` gains an optional `decimals` arg (default `1`, behavior-preserving for every existing caller, including the #444 999950 boundary); the bar calls `formatTokens(n, 0)`, keeping its compact whole-number `k` and gaining the rollup, e.g. `711k/1.0M`.

Tested: `bun run lint` clean; `bun test` green except the same Windows-environment failures `main` has.

### DIFF
--- a/src/utils/__tests__/renderer-format-tokens.test.ts
+++ b/src/utils/__tests__/renderer-format-tokens.test.ts
@@ -33,4 +33,11 @@ describe('formatTokens', () => {
     it('still renders just below the boundary as k', () => {
         expect(formatTokens(999949)).toBe('999.9k');
     });
+
+    it('uses whole-number k and rolls up to M at decimals=0', () => {
+        expect(formatTokens(711000, 0)).toBe('711k');
+        expect(formatTokens(999499, 0)).toBe('999k');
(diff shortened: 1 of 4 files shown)


zachthedev:
The Windows-environment test failures noted in the description are fixed separately in #454; with that branch, `bun test` is fully green on Windows too.

Issue #450, opened 2026-06-15
compaction-counter: expose count/auto/manual/unknown/reclaimed as selectable raw values for composition


zachthedev:
`compaction-counter` renders count + trigger split + reclaimed as one fixed string (`N (n auto, n manual) <arrow>X`). To let a statusline compose its own layout (own icons/separators between parts, like the token widgets do), add a metadata selector for which sub-metric the widget outputs — e.g. `metric: count|auto|manual|unknown|reclaimed` — emitting that single number in `rawValue` mode.

Users could then concatenate several instances with custom-symbol labels/separators instead of the built-in format. The current combined output stays the default.

PR #431, opened 2026-06-13
feat: allow overriding widget symbols with a custom glyph

zachthedev:
Closes #436.

Lets users replace or remove any widget's hardcoded symbol.

`item.character` already worked on Git Staged/Unstaged/Untracked but had no editor. This extends it to every widget with a hardcoded glyph and adds a shared `(g)lyph` keybind plus a tiny editor: type a character or emoji, Backspace renders the widget without a symbol, picking the default clears the override.

| Widget                            | Default                                   |
| --------------------------------- | ----------------------------------------- |
| Git Branch                        | `⎇`                                       |
| Git Worktree                      | `𖠰`                                       |
| Git Worktree Mode                 | `⎇`                                       |
| Git Conflicts                     | `⚠`                                      |
| JJ Bookmarks                      | `🔖`                                      |
| JJ Workspace                      | `◆`                                       |
| Git Staged / Unstaged / Untracked | `+` / `*` / `?` (now editable in the TUI) |

12 files, +366/-33. New `shared/symbol-override.tsx` carries the keybind, helpers, and editor; each widget adds ~15 lines (keybind entry, `renderEditor` delegation, prefix helper in `render`). Preview-path tests cover default/override/suppressed for all nine widgets, plus helper units.

Deferred: Git Ahead/Behind (paired `↑`/`↓`) and Git Status (composite output) need a different shape than the single `character` field; happy to follow up if there's interest.

Motivation: nerd-font users who pair widgets with custom-symbol icons end up with doubled glyphs (` ⎇ main`); overriding or blanking the builtin fixes that without touching the renderer.

Note: #430 touches some of the same widget files; whichever lands second gets a trivial rebase (conflicts are line-local).

Tested: `bun run lint` clean; `bun test` green except the same Windows-environment failures `main` has.

PR #452, opened 2026-06-15
feat(compaction): make the reclaimed-tokens arrow an overridable glyph

### LINKED ISSUE #448: compaction-counter: hardcoded reclaimed-tokens arrow is not overridable like other widget symbols
`formatReclaimedSuffix` in `src/widgets/CompactionCounter.ts` hardcodes a leading down-arrow (U+2193) before the reclaimed-token count shown by `showReclaimed`. After #431 added per-widget symbol overrides, this is the one widget glyph with no `character`/slot override path.

Request: expose the reclaimed prefix as an overridable symbol slot (default U+2193), consistent with the git/jj widget symbols.

Secondary: `formatTriggerSuffix` wraps the trigger split in parentheses; a separator-based layout would read more consistently next to other widgets.

zachthedev:
Closes #448

`formatReclaimedSuffix` hardcoded a leading ↓ before the reclaimed-token count.
#431 gave the git/jj widgets per-symbol overrides but skipped this one.

Routes the reclaimed prefix through the existing symbol-override slot system as a
`symbolReclaimed` metadata slot (default ↓), reachable via the shared `(g)lyph`
editor and `renderSymbolSlotsEditor`.

- Default unchanged: existing configs still render ` ↓1.2M`.
- Custom glyph (e.g. `symbolReclaimed` = `-`) renders ` -1.2M`.
- Empty override drops the glyph but keeps the separating space (` 1.2M`).

The secondary parens-vs-separator suggestion in #448 is layout/composability and
fits #450 better; left out of scope here.

Tests: added custom-glyph and empty-override cases and updated the keybind
assertions for the new `(g)lyph` bind. `bun run lint` clean; `bun test` green
except the pre-existing Windows-env failures (terminal stty/tput, non-macOS
usage-token, buildCommand path-quoting) that also fail on `main`.

### DIFF
--- a/src/widgets/CompactionCounter.ts
+++ b/src/widgets/CompactionCounter.ts
@@ -7,6 +7,7 @@ import type {
     CustomKeybind,
     Widget,
     WidgetEditorDisplay,
+    WidgetEditorProps,
     WidgetItem
 } from '../types/Widget';
 import { ZERO_COMPACTION_STATS } from '../utils/compaction';
@@ -16,6 +17,12 @@ import {
     isMetadataFlagEnabled,
     toggleMetadataFlag
 } from './shared/metadata';
+import {
+    getSlotSymbol,
+    getSymbolKeybind,
+    renderSymbolSlotsEditor,
+    type SymbolSlot
+} from './shared/symbol-override';
 
 const COMPACTION_ICON = '↻';
 const COMPACTION_NERD_FONT_ICON = '\uF021';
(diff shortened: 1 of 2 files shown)


zachthedev:
The pre-existing Windows-env test failures noted in the description are fixed separately in #454; with that branch, `bun test` is fully green on Windows too.

CorticalCode:
Thanks for the tweak and the #454 fix 👍

PR #456, opened 2026-06-15
feat: configurable numeric precision (per-widget, type-aware, optional global)

### LINKED ISSUE #455: feat: configurable numeric precision (per-widget, type-aware, optional global)
## Problem

The numeric formatters render a fixed decimal count, so whole values carry a
trailing `.0`: `1.0M`, `512.0k`, `100.0%`, `50.0 t/s`. Blanket-stripping it
(the original #456) was rejected: the fixed width stops the field from snapping
as a value crosses a whole unit (`1M` -> `1.1M` -> ...), which Claude Code
itself preserves (`6.0k`). The fix is to make precision configurable, not to
change the default.

## Proposal

A per-widget number-format setting, type-aware, with an optional global.
Default output is unchanged.

### Format: `numberFormat = { style?, decimals? }`

- `style`: `precise` (keep trailing zeros, today: `1.0M`) | `compact` (drop
  trailing zeros, keep real fractions: `1M`, `1.1M`) | `whole` (no decimals: `1M`)
- `decimals`: max places before `style` applies (`precise` + 2 -> `1.00M`)

### Type-aware defaults

Each numeric widget has a number _kind_ with its own default, so one setting
does the right thing per type:

| kind    | widgets                                                                                               | default                 |
| ------- | ----------------------------------------------------------------------------------------------------- | ----------------------- |
| token   | tokens in/out/cached/total, context length, context-bar count, compaction reclaimed, cache read/write | precise, 1 (`1.0M`)     |
| speed   | input/output/total speed                                                                              | precise, 1 (`50.0 t/s`) |
| percent | context %, cache hit %, usage/utilization/reset sliders                                               | precise, 1 (`84.5%`)    |
| memory  | free memory                                                                                           | current (G: 1, M/K: 0)  |
| cost    | session cost                                                                                          | precise, 2 (`$1.20`)    |

So tokens can be `xxx.x` while money stays `xxx.xx`: cost defaults to 2 and a
token-oriented change won't pull it to 1.

### Precedence

`effective = global[kind] (if set) -> widget.numberFormat (if set) -> kind default`

Per-widget is the primary control. The global is optional and off by default;
when set it wins, the same way `overrideForegroundColor` / `globalBold` already
do. With nothing set, every widget renders exactly as today.

## Scope

`formatTokens` + consumers, `formatSpeed`, `FreeMemory`, the `toFixed(1)%`
percentage sites, `SessionCost`.

## Implementation notes

- `numberFormat` is an optional top-level `WidgetItem` field (parallel to
  `dim`/`bold`); global is an optional per-kind `Settings` field. Optional, so
  no settings-version migration; absent = current behavior.
- One `resolveNumberFormat(kind, item, settings)` choke point; pure formatters
  take the resolved `{ style, decimals }`; shared helpers take it as a param.
  Precision is resolved at render time (it can't be a post-render pass like
  color/bold).
- TUI: per-widget affordance in the editor (like `(d)` for dim) + a global
  entry in the overrides menu.

## Non-goals

- No change to default output. Cost stays 2-decimal by default.

zachthedev:
Closes #455

Replaces the rejected "always trim `.0`" approach (per the discussion here) with
configurable decimal precision for numeric widgets. Default output is unchanged.

## Model

- Per-widget `numberFormat: { style?, decimals? }` (top-level `WidgetItem` field,
  alongside `dim`/`bold`):
  - `style`: `precise` (keep trailing zeros, today's `1.0M`) | `compact` (trim
    them: `1M`, `1.1M`) | `whole` (no decimals: `1M`).
  - `decimals`: max places before `style` applies.
- Optional per-kind global `Settings.numberFormat` (`token`, `speed`, `percent`,
  `memory`, `cost`). A kind set there wins over the per-widget value, same
  precedence as `overrideForegroundColor` / `globalBold`.
- Type-aware baselines, so a token-oriented change never pulls money off two
  decimals: token/speed/percent default 1, cost 2, memory G:1 / M:0 / K:0.
- `resolveNumberFormat(kind, item, settings)` is the single choke point; the
  pure formatters (`formatTokens`, `formatSpeed`, `formatPercent`, `formatCost`,
  memory) take the resolved format. Optional fields, so no settings migration.

## Scope

Every numeric widget: tokens in/out/cached/total, context length/window/bar,
compaction reclaimed, cache read/write/hit, free memory, session cost,
input/output/total speed, context %/usable, the usage sliders (session, weekly,
weekly opus/sonnet, weekly fable, extra-usage), and the block/weekly timers.

## TUI

Per the review feedback, the per-widget control lives on the line editor rather
than the color menu, so every override other than color stays on one screen and
remains reachable while a powerline theme is active.

- Line editor: `(.)` cycles the highlighted widget's style (default -> compact ->
  whole). A widget opts in with `supportsNumberFormat()`, matching the existing
  `supportsRawValue()` / `supportsColors()` convention, and the editor injects
  the keybind for any widget that returns true. `.` reads as the decimal point
  and leaves every letter free for widget-specific binds.
- Global overrides: `(n)` opens a per-type submenu to cycle each kind's style.
- `decimals` is set per-widget / per-kind in `settings.json` (the power-user
  axis); the TUI exposes the style.

## Tests

`bun run lint` clean; `bun test` green (1896 pass, plus the one
`global-command-resolution` failure `main` already has on this host). Default
output is unchanged (every existing widget/formatter assertion still passes);
added coverage for the format primitives, resolver precedence, the style cycle,
and the line editor's injected keybind.

Rebased onto current `main`. The CompactionCounter reconcile with #457 is
included: the metric selector's `reclaimed` value now honors the resolved
precision.


CorticalCode:
@zachthedev 

Hi, I wanted to advocate for leaving the decimal point in place. At first glance I figured yeah, that trailing zero shouldn't be there. But thinking about it, on a live status line it's actually doing a job: it holds the field width steady as the number climbs. Drop it and the width snaps every time the count crosses a whole unit (`1M`, then `1.1M`, and so on), which ends up more distracting than the `.0` it removes.

Claude Code keeps follows this behavior in its token count: past `999` the decimal is preserved through the zeros (e.g. `6.0k`, not `6k`). 

Screenshot 2026-06-15 at 10 01 25 PM

zachthedev:
> @zachthedev 
> 
> Hi, I wanted to advocate for leaving the decimal point in place. At first glance I figured yeah, that trailing zero shouldn't be there. But thinking about it, on a live status line it's actually doing a job: it holds the field width steady as the number climbs. Drop it and the width snaps every time the count crosses a whole unit (`1M`, then `1.1M`, and so on), which ends up more distracting than the `.0` it removes.
> 
> Claude Code keeps it too  in its token counter: past `999` the decimal is preservedthrough the zeros (e.g. `6.0k`, not `6k`). 
> 
> Screenshot 2026-06-15 at 10 01 25 PM
> 

That is a good point, it should be configurable honestly. I am going to rework it to be a per element or global config to have decimals or not. Instead of chopping off .0.

sirmalloc:
@zachthedev i prefer this be a widget specific config, not global.

zachthedev:
> @zachthedev i prefer this be a widget specific config, not global.

What about widget specific with a global override? If a user didn't want to manually configure every one.
Default would be as it is nowadays.

sirmalloc:
> > @zachthedev i prefer this be a widget specific config, not global.
> 
> What about widget specific with a global override? If a user didn't want to manually configure every one. Default would be as it is nowadays.

What widgets would it impact?  Anything displaying a number?  It'll touch more widgets but if it makes sense I'm fine with it if it's done cleanly and doesn't modify existing default behavior.

zachthedev:
> > > @zachthedev i prefer this be a widget specific config, not global.
> > 
> > What about widget specific with a global override? If a user didn't want to manually configure every one. Default would be as it is nowadays.
> 
> What widgets would it impact?  Anything displaying a number?  It'll touch more widgets but if it makes sense I'm fine with it if it's done cleanly and doesn't modify existing default behavior.

Yeah anything displaying a number. I marked this as a draft and will mess around with a few ideas.

sirmalloc:
@zachthedev I'm not sure the color menu is the proper place to manage this, it makes more sense on the Widget Line editor.  I tried editing colors on my current config and as the code is now if i have a powerline theme active I can't edit colors.  This may be changing in the future with some other plans but for now all widget-related overrides other than color have typically been handled on the line editor screen.

zachthedev:
Moved it off the color menu, and rebased onto current `main`.

The per-widget control now lives on the line editor. A widget opts in with
`supportsNumberFormat()`, matching the existing `supportsRawValue()` /
`supportsColors()` convention, and the items editor injects the keybind for any
widget that returns true, so no numeric widget needs keybind wiring of its own.
That also fixes the case you hit: it no longer matters that the color menu is
unreachable with a powerline theme active.

The keybind is `(.)`, which reads as the decimal point. Every letter was already
taken basically.

The color menu is back to color, bold and dim only: the `(n)` handler, the
`[#style]` indicator, and the footer hint are all gone, and the cycle moved to
`utils/number-format.ts` so the number-format logic sits in one module.

Two other things I did:

- Weekly Fable Usage landed since this branch opened and was rendering
  hardcoded `toFixed(1)`, so it routes through `formatPercent` now.
- The #457 CompactionCounter reconcile, as flagged there: the metric selector's
  `reclaimed` value honors the resolved precision.

PR #458, opened 2026-06-16
feat(tui): warn on an invalid settings.json and confirm before overwriting it

CorticalCode:
## Summary

Follow-up to #447. That PR made ccstatusline recover non-destructively from an invalid/unreadable `settings.json` and show a red `⚠ invalid config` badge on the **piped** status line, recording the reason via `getConfigLoadError()`. The **interactive configurator** never used that signal: it silently fell back to defaults, and saving from the TUI would overwrite the preserved-but-broken file — the exact data loss #447 set out to prevent.

This wires the existing signal into the TUI: a persistent warning when the config was rejected, plus a one-time confirmation before any TUI save can overwrite the preserved file.

## The gap after #447

On an unreadable `settings.json`:
- ✅ **piped render** → shows `⚠ invalid config`, file left intact
- ❌ **TUI** → silently shows **default** settings with no indication the file was rejected; a subsequent **Ctrl+S** or **Save & Exit** calls `saveSettings()` and overwrites the recoverable original

## What this changes

On launch the TUI reads `getConfigLoadError()` once into component state (the module signal is volatile). When it's set:

1. **A persistent red banner** under the header names the reason and warns that saving overwrites the file:
   ```
   ⚠ settings.json is not valid JSON — showing defaults; saving here overwrites the file.
   ```
2. **Both explicit-save routes are guarded** — Ctrl+S and "💾 Save & Exit" route through a one-time confirm:
   ```
   settings.json is not valid JSON and is preserved on disk. Saving replaces it with the current configuration. Continue?
   ```
   Confirm → save (banner/guard then clear, since the file is valid again). Cancel → file untouched, banner stays.

The guard is a single confirmation, **not a lock** — a user who wants to replace the file just confirms once. No `.bak`/backup files: the protection is consent plus leaving the file intact until the user chooses to save.

| Scenario | Before | After |
|---|---|---|
| Valid `settings.json` | normal | unchanged |
| Invalid/unreadable config on open | silently shows defaults | persistent red banner naming the reason |
| Ctrl+S / Save & Exit, valid config | saves | saves (unchanged) |
| Ctrl+S / Save & Exit, **invalid** config | **silently overwrites the preserved file** | confirm first → save; cancel → file untouched |
| After a confirmed save | — | banner clears; guard doesn't fire again |

## Notes

- **Gate, not a new writer.** This only wraps the existing `saveSettings()`, so it composes with the symlink-preserving write from `b20296c` — saves still resolve symlinks. The installation-metadata writer is already guarded by #447's internal check; after install/update/uninstall (which re-read settings internally) the captured signal is re-synced so the banner can't go stale.
- The confirm message reflects the **actual** reason (`is not valid JSON` / `is not in a valid format` / `could not be read`), not a fixed string.
- Logic lives in two small pure helpers (`buildConfigLoadWarning`, `buildInvalidConfigSaveConfirm`) unit-tested in `App.test.ts`, matching the existing "test the helper, not the Ink tree" pattern.
- The banner is intentionally omitted from the pinned-version-mismatch screen (no save is reachable there).

## Testing

- New unit tests for both helpers (null-guard, banner content, confirm shape, reason interpolation, `onConfirm` wiring).
- Full suite green (1557 tests), `bun run lint` clean.
- Manual: launch against an invalid file (`bun run src/ccstatusline.ts --config <tmp>/settings.json`) → banner shows; Ctrl+S → confirm → **cancel** leaves the file untouched, **confirm** saves and clears the banner.

Happy to trim to **banner-only** (dropping the save-guard) if you'd prefer a lighter touch.


sirmalloc:
Thanks, it'll go out in the next release.

PR #473, opened 2026-06-20
fix: invert plain usage percentages

### LINKED ISSUE #391: Session/Weekly usage widgets lack a (u)sed/remaining toggle, and existing (v)invert is silently dropped in plain mode
### Summary

Two related problems with the inversion behavior on the percentage-style usage widgets:

1. **No `(u)sed/remaining` toggle.** `Context % (usable)` exposes a discoverable `u` keybind labeled "(u)sed/remaining" that flips the displayed value to `100 − percent`. The sibling usage widgets (Session, Weekly, Weekly Opus, Weekly Sonnet, Extra Usage Utilization) use a different keybind — `v` labeled "in(v)ert fill" — for the same operation. Two issues with this:
   - "(v)ert fill" reads like a visual styling option, not a semantic "show remaining instead of used" toggle.
   - Inconsistent with the established `u` convention on the context widget.

2. **The existing `v` invert is silently dropped in plain-percentage mode.** In all 5 affected widgets, the live render computes:

   ```ts
   const percent = Math.max(0, Math.min(100, data.sessionUsage));
   const renderedPercent = inverted ? 100 - percent : percent;
   ```

   but then the plain-percent fallback path renders `percent.toFixed(1)` instead of `renderedPercent.toFixed(1)`. So pressing `v` only affects progress-bar and slider display modes — in the default plain-percent mode (which is what most raw-value layouts use), the toggle silently does nothing. Same bug in the preview path (`previewPercent` instead of `renderedPercent`).

### Affected files

| File | Live-mode line | Preview-mode line |
|---|---|---|
| `src/widgets/SessionUsage.ts` | 117 | 82 |
| `src/widgets/WeeklyUsage.ts` | 117 | 82 |
| `src/widgets/WeeklyOpusUsage.ts` | 119 | 84 |
| `src/widgets/WeeklySonnetUsage.ts` | 119 | 84 |
| `src/widgets/ExtraUsageUtilization.ts` | 112 | 82 |

### Repro

1. Add a `Session Usage (raw value)` widget to your statusline (no progress, no slider).
2. Open the TUI editor (`ccstatusline`), select the widget, press `v` ("invert fill").
3. Observe: the displayed percentage doesn't change. There's no visible indicator that anything happened.

### Suggested fix

- Rename / alias the keybind from `v` "(v)ert fill" → `u` "(u)sed/remaining" in `src/widgets/shared/usage-display.ts`. Keep `v` as a back-compat alias so muscle memory doesn't break.
- In each of the 5 widgets, swap `percent.toFixed(1)` → `renderedPercent.toFixed(1)` on the plain-percent live render path, and `previewPercent.toFixed(1)` → `renderedPercent.toFixed(1)` on the preview path.

PR coming.

ekkoitac:
Fixes #391

## Summary

This makes the percentage usage widgets treat used/remaining as a regular display option instead of only a progress-bar setting.

- Add `u` for `(u)sed/remaining` on the percentage usage widgets.
- Keep the existing `v` shortcut working.
- Use the inverted value in plain text and preview mode for Session, Weekly, Weekly Opus, Weekly Sonnet, and Extra Usage Utilization.
- Keep `invert` when cycling back to plain text mode, while still dropping cursor metadata since the cursor only applies to bars and sliders.

## Validation

- `npm exec --yes bun@latest -- test src/widgets/__tests__/SessionUsage.test.ts src/widgets/__tests__/WeeklyUsage.test.ts src/widgets/__tests__/WeeklyOpusUsage.test.ts src/widgets/__tests__/WeeklySonnetUsage.test.ts src/widgets/__tests__/ExtraUsageUtilization.test.ts`
- `npm exec --yes bun@latest -- test`
- `npm exec --yes bun@latest -- run lint`


sirmalloc:
I made a couple small tweaks.  I got rid of the 'v' fallback to avoid confusion.  I make the widget helper text show whether we are in used or remaining mode all the time now to eliminate any lingering confusion.  And I updated the docs.  Thanks for the contribution!

PR #510, opened 2026-07-17
fix: read per-model weekly usage from limits[]; derive wiring from a single registry

elhoim:
## Summary

A raw `curl` against `/api/oauth/usage` (2026-07) shows the real response shape has moved on from what this code assumed:

```json
"seven_day_sonnet": null,
"seven_day_opus": null,
"limits": [
  { "kind": "session", "percent": 48, ... },
  { "kind": "weekly_all", "percent": 18, ... },
  { "kind": "weekly_scoped", "percent": 3, "scope": { "model": { "display_name": "Sonnet" } }, ... }
]
```

The legacy flat seven_day_sonnet/seven_day_opus buckets return null even for a model with real usage that week — the real per-model number now lives in a limits[] array, matched by scope.model.display_name.

- parseUsageApiResponse now looks up each model's weekly_scoped limit first, falling back to the legacy flat bucket only when no such entry exists (preserving the existing null → 0% convention for that fallback, for accounts/API versions where the flat field is still authoritative).

This also fixed the underlying wiring the fallback path depends on: adding a per-model weekly usage bucket (Sonnet, Opus) previously required hand-editing four separate lookup tables in usage-prefetch.ts (USAGE_WIDGET_TYPES, USAGE_DATA_FIELDS, USAGE_WIDGET_REQUIREMENTS, USAGE_CURSOR_REQUIREMENTS) plus WINDOW_RESET_FIELD_SENTINELS in usage-fetch.ts. None of those tables were compiler-linked, so omitting a bucket from just one of them compiled cleanly and failed silently at runtime (e.g. the widget always renders null, or the progress cursor silently falls back to the wrong reset window).

- WEEKLY_MODEL_USAGE_BUCKETS (usage-types.ts) is now the single source of truth those tables (and the limits[] lookup above) are derived from.
- The one remaining hand-duplicated spot — CachedUsageDataSchema and UsageApiResponseSchema in usage-fetch.ts both need a model's fields declared by hand, since Zod needs statically-known object keys to keep field-level type inference for the parse* functions — is now covered by a schema-parity test (usage-fetch.test.ts) asserting both schemas cover every WEEKLY_MODEL_USAGE_BUCKETS entry, plus round-trip tests simulating an API fetch followed by a cache write/read. A bucket added to one schema but not the other now fails a test instead of silently vanishing after the next cache round-trip.

Test plan

- [x] bun run lint passes
- [x] bun test — 1648/1650 pass; the 2 failures (a sandboxed subprocess-timeout probe test, an Ink keypress-timing test) are pre-existing environmental flakiness unrelated to this change
- [x] New tests use the exact JSON shape captured from a live response
- [x] Manual CLI check: seven_day_sonnet: null + no limits[] still renders Weekly Sonnet: 33.0% / Weekly Opus: 0.0% unchanged (legacy fallback preserved); a weekly_scoped limits[] entry is correctly preferred over a stale/null legacy bucket for the same model

### DIFF
--- a/src/types/StatusJSON.ts
+++ b/src/types/StatusJSON.ts
@@ -19,6 +19,8 @@ const RateLimitPeriodSchema = z.object({
     resets_at: CoercedNumberSchema.nullable().optional() // Unix epoch seconds
 });
 
+export type RateLimitPeriod = z.infer<typeof RateLimitPeriodSchema>;
+
 export const StatusJSONSchema = z.looseObject({
     hook_event_name: z.string().optional(),
     session_id: z.string().optional(),
--- a/src/utils/__tests__/usage-fetch.test.ts
+++ b/src/utils/__tests__/usage-fetch.test.ts
@@ -11,6 +11,9 @@ import {
     it
 } from 'vitest';
 
+import { __testing } from '../usage-fetch';
+import { WEEKLY_MODEL_USAGE_BUCKETS } from '../usage-types';
+
 const require = createRequire(import.meta.url);
 const { execFileSync: realExecFileSync } = require('node:child_process') as { execFileSync: typeof childProcess.execFileSync };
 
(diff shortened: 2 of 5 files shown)


sirmalloc:
Squashed this to main with a few fixes, but wasn't able to push the fixes to your PR branch because it was protected.  Because of this I had to squash directly to main, which is why this shows as Closed instead of Merged.  But the commit does have you listed as co-author for contributor credit.  Thanks for the PR!

PR #548, opened 2026-08-11
fix: bound stdin read so the process cannot outlive its render

filipemb:
## Problem

`readStdin()` iterates `process.stdin` until EOF. When the host writes the status JSON but keeps the write end open, that iteration never settles: the render finishes, but the process stays resident. On a long-lived host with a short refresh interval these accumulate — I found 77 of them alive, the oldest at 257 hours, holding ~2.3 GB.

Repro — spawn the CLI, write the payload, never call `end()`:

```js
const p = spawn(process.execPath, ['dist/ccstatusline.js'], { stdio: ['pipe', 'pipe', 'pipe'] });
p.stdin.write(JSON.stringify({ session_id: 't', model: { display_name: 'Opus' } })); // no end()
setTimeout(() => console.log('still alive'), 12000);
```

Before: still alive, indefinitely. After: exits in ~5s, having rendered the line.

## Fix

Race the read against a timeout — 5000 ms default, `CCSTATUSLINE_STDIN_TIMEOUT_MS` to override. On timeout it resolves with whatever already arrived instead of discarding it: the payload is written in one shot, so a missing EOF means the EOF is missing, not the data, and the status line still renders correctly. `finally` clears the timer and pauses/unrefs the stream.

`readStdin` moves to `src/utils/stdin.ts` because the entrypoint runs `main()` on import, which makes the function untestable where it was.

## Tests

`src/utils/__tests__/stdin.test.ts` covers three cases: TTY returns null, a normal close reads the payload, and no-EOF resolves with what arrived. The third one fails on `main` (times out) and passes with this change. The rest of the suite shows no new failures.

### DIFF
--- a/src/ccstatusline.ts
+++ b/src/ccstatusline.ts
@@ -46,6 +46,7 @@ import {
     getWidgetSpeedWindowSeconds,
     isWidgetSpeedWindowEnabled
 } from './utils/speed-window';
+import { readStdin } from './utils/stdin';
 import {
     getPackageVersion,
     getTerminalWidth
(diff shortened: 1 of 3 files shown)


filipemb:
Rebased onto current main (2522c11) and pushed the Node-compatible stdin fix plus focused regression tests. Local lint and build pass; the stdin tests are 3/3, the bundled CLI exits normally with stdin held open, and the full suite reached 1870 passes with one unrelated flaky UsageTimezoneEditor case that passed 4/4 in isolation. GitHub CI is currently action_required pending maintainer approval for the fork workflow. @sirmalloc, this is ready for another look.

PR #553, opened 2026-08-15
feat: Claude Status widget with incident history strip

JimmySpivey:
## What

A new `claude-status` widget showing live Claude/Anthropic service health from status.claude.com, with an optional incident-history strip.

- **Status word** from `/api/v2/status.json` — `ok` / `minor` / `major` / `critical` / `maintenance`, severity-colored; degrades to `?` on fetch/parse failure without breaking the line.
- **History strip** (toggle with `h` in the editor): the past 48h as 8 six-hour buckets, one `▮` each, colored by the worst incident whose `created_at → resolved_at` window overlaps the bucket (green none / yellow minor / orange major / red critical). Built from `/api/v2/incidents.json`, which is only fetched when some item has history enabled.

## How

- Fetch/cache layer (`src/utils/claude-service-status.ts`) modeled on the existing usage prefetch: async prefetch wired next to usage prefetch in `ccstatusline.ts`, disk cache in `~/.cache/ccstatusline/` with 5-min TTL, 30s failure backoff, stale-cache fallback, HTTPS_PROXY support, zod-validated parsing.
- The multi-colored strip needed the renderer's preserve-colors path, which was hardcoded to `custom-command`. Generalized it into an optional `preservesRenderedColors()` on the Widget interface (implemented by CustomCommand — behavior unchanged — and ClaudeStatus in history mode). Bars reset foreground only (`ESC[39m`) so powerline backgrounds survive.
- Orange has no ansi16 slot: major maps to red there, ansi256 208 / truecolor `#ff8700` otherwise. `colorLevel: 0` emits plain text.

## Testing

- 24 new tests (bucket overlap incl. boundary/unresolved incidents, impact ranking, malformed JSON, degraded render, editor toggle, preserve-colors flags) — no network, all mocked.
- `bun run lint` and `bun run build` pass; `bun test` 1891 pass with the same single pre-existing flake (`global-command-resolution`) that fails on pristine main.
- Live smoke against the real endpoints rendered current status + the actual Aug 13–15 incidents in the strip.


JimmySpivey:
image - each bar is the last 6 hours, 8 bars, 48 hours.
image - colors, red - major outage, yellow - partial outage, yellow green - minor incident, green - operational with no issues

